PR #29, opened 2026-07-03
fix: 7 defects found in weekly review (bundle TOCTOU, redirect 500, Python/Dart SDK bugs)

DennisAlund:
Weekly defect-hunting review of the app, API, and all three SDKs. Seven confirmed bugs fixed, each with a regression test. Two items flagged as PR comments for developer judgment (not committed).

All 1047 app tests pass. All 94 Python SDK tests pass. Dart SDK tests written but not verified locally (no `dart` binary in CI environment).

---

## Fix 1: Bundle service TOCTOU — null/false return values not checked

**File:** `src/services/bundle-management.ts`

`deleteBundle` discarded the `boolean` returned by `BundleRepository.delete`, always calling `return ok({ deleted: true })` even when no row was found. `updateBundle`, `archiveBundle`, and `unarchiveBundle` all used non-null assertions (`!`) on repository returns that become `null` when a concurrent DELETE races between the fetch and the write. Each now captures the return value and calls `return fail(404, "Bundle not found")` on `null`/`false`.

**Impact:** A request to delete/update/archive/unarchive a bundle that was concurrently deleted returns `ok` with a stale or invented body instead of `404`. Under non-concurrent conditions the assertions are vacuously safe, but a simple race between two admin users triggers them.

**Regression tests:** Four new tests in `src/__tests__/service/bundle-service.test.ts`, each using `vi.spyOn(BundleRepository, "<method>").mockResolvedValueOnce(false/null)` to simulate the race.

---

## Fix 2: Redirect handler — unguarded `new URL()` crashes with 500

**File:** `src/redirect.ts`

The redirect handler called `new URL(entry.url)` without a `try/catch`. A corrupted or legacy KV entry with a non-parseable URL string throws a `TypeError` that was uncaught, surfacing as an unhandled 500. Wrapped in `try/catch`; returns `notFoundResponse()` on parse failure.

**Regression test:** `redirect-kv.test.ts` — "malformed URL in KV cache returns 404 instead of 500".

---

## Fix 3: Python SDK — `parse_json_response` does not wrap `JSONDecodeError` on 2xx

**File:** `sdk/python/src/shrtnr/_base.py`

`parse_json_response` returned `response.json()` bare on success paths. An HTML error page or truncated body on a 200 (CDN edge, misconfigured proxy) raised a raw `JSONDecodeError` instead of a typed `ShrtnrError`. Wrapped in `try/except`; raises `ShrtnrError(response.status_code, ...)` so callers see a consistent exception type.

**Regression test:** `test_non_json_2xx_raises_shrtnr_error` — mocks a 200 returning `<html>Bad Gateway</html>` and asserts `ShrtnrError` is raised with `status == 200`.

---

## Fix 4: Python SDK — `LinksResource.qr()` `size` parameter typed `str` instead of `int`

**File:** `sdk/python/src/shrtnr/resources/links.py`

Both `LinksResource.qr()` (sync, line 174) and `AsyncLinksResource.qr()` (async, line 330) declared `size: str | None`. The OpenAPI spec defines `size` as an integer; the TypeScript and Dart SDKs both accept `int`. Passing an integer worked accidentally (str conversion via query encoding), but type checkers flagged it and the docstring was wrong. Changed to `int | None`; `str(size)` applied when building the query param.

**Regression tests:** `test_links_qr_size_accepts_int` and `test_links_qr_size_omitted_when_none`.

---

## Fix 5: Dart SDK — `Bundle.fromJson` / `BundleWithSummary.fromJson` crash on null `accent`

**File:** `sdk/dart/lib/src/models.dart`

Both factories cast `json['accent'] as String` (non-nullable). When a bundle was created before the accent field existed, or when the API omits it (nullable in schema), the cast throws a `TypeError` at runtime. Changed to `(json['accent'] as String?) ?? 'orange'` to match the server-side default.

**Regression tests:** Four new tests in `sdk/dart/test/client_test.dart` — `Bundle.fromJson` and `BundleWithSummary.fromJson`, each with accent absent and accent explicitly null, both expect `BundleAccent.orange`.

---

## Flagged items (not committed — need developer judgment)

These were noted during review but not fixed. Added as PR review comments on the relevant lines.

**1. JWT issuer not validated (`src/access.ts`)** — `jwtVerify` is called without an `issuer` option. A token signed by a different Cloudflare Access application (same account, different `aud`) passes the audience check but should be rejected. Fix requires knowing the expected issuer URL for each deployment, so left for the developer.

**2. `slugClickCountSql` interpolates `sinceTs` directly into SQL (`src/db/filters.ts`)** — The `sinceTs` value comes from a controlled enum translation so injection is low-probability, but it is a raw string interpolation. A parameterized query would eliminate the risk without loss of readability.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LMdebR2GAyPam7cGgnTeM3)_


DennisAlund:
**Flagged (not fixed): JWT issuer not validated — `src/access.ts`**

`jwtVerify` is called without an `issuer` option in both the `cfAccess` and any future JWKS-backed verifier paths. Cloudflare Access tokens carry an `iss` claim of the form `https://<team>.cloudflareaccess.com`. Without checking it, a valid JWT issued by a *different* CF Access application on the same account (different `aud`, same signing key infrastructure) would pass the audience check after a misconfiguration.

The fix is straightforward — add `issuer: "https://<team>.cloudflareaccess.com"` to the `jwtVerify` options — but it requires knowing (or injecting via env) the team domain for each deployment. I left it unfixed to avoid hardcoding a value I can't verify from the codebase alone.

Recommended action: add `ACCESS_ISSUER` to the env vars alongside `ACCESS_AUD` and pass it to `jwtVerify`.

---

DennisAlund:
**Flagged (not fixed): raw string interpolation in `slugClickCountSql` — `src/db/filters.ts`**

`sinceTs` is interpolated directly into the SQL string rather than passed as a bound parameter:

```ts
// src/db/filters.ts ~line 47
`... WHERE click_ts >= ${sinceTs} ...`
```

The value comes from a controlled enum translation so actual injection is low-probability, but a bound parameter would eliminate the risk with no downside. Worth a one-line fix when the file is next touched.

---

PR #40, opened 2026-08-21
Weekly defect review: i18n interpolation, range localization, bundle stats, Python SDK redirects

DennisAlund:
Weekly automated defect-hunting review across the app, API, and all three SDKs. Four fixes qualified (unambiguous, local, low-risk, test-covered); everything else is left as a PR comment for a judgment call.

### 1. i18n interpolation corrupted by `String.replace` special patterns
`createTranslateFn` substituted `{param}` placeholders with `value.replace(regex, String(v))`. A *string* replacement argument treats `$&`, `` $` ``, `$'`, `$1` etc. as special patterns, not literal text. Any interpolated value containing one of those sequences — an API key title, a bundle name, a link URL, all user-controlled free text — silently corrupted the rendered confirm-dialog/toast copy.

**Fix:** pass a replacer function so the value always splices in literally. **Test:** `src/__tests__/unit/i18n.test.ts` reproduces the corruption with a `` "x$`y" `` value.

### 2. Clicks column header showed the raw range key untranslated
The links page's "Clicks ({range})" header substituted the raw internal range key (`"24h"`) instead of going through `t()`, unlike every other range display in the codebase. Cosmetic in English, wrong in `id`/`sv` (showed `"24h"` instead of `"24J"`/`"24T"`).

**Fix:** route it through `t(`range.${range}`)`, matching the pattern used elsewhere. **Test:** requests the links page in the `id` locale and asserts the localized label appears.

### 3. Bundle `top_country.pct` computed against the wrong denominator
`getBundleStats`'s countries list is capped at `LIMIT 10`; `top_country.pct` divided the leading country's count by the sum of only that top-10 list instead of the bundle's true `total_clicks`. Once a bundle spans more than 10 distinct countries, the pct is inflated. The sibling field `per_link[].pct_of_bundle` already uses the correct denominator a few lines above.

**Fix:** use `totalClicks` for both fields. **Test:** seeds 11 distinct countries (one dominant, ten tied at the LIMIT-10 cutoff) so the two denominators diverge, and asserts the corrected value.

### 4. Python SDK didn't follow redirects
`httpx.Client`/`AsyncClient` default to `follow_redirects=False`. Neither `Shrtnr` nor `AsyncShrtnr` overrode that on the client they own. The TypeScript SDK (`fetch`, default "follow") and Dart SDK (`http.Client`, follows by default) both handle a 3xx from a deployment's front door transparently; Python raised `ShrtnrError` for the identical request. Single-SDK fix: TS/Dart already behave the way this brings Python in line with.

**Fix:** pass `follow_redirects=True` when constructing the owned client. **Tests:** sync and async tests mock a 301→200 chain and assert the request completes.

All four regression tests were confirmed to fail on the pre-fix code and pass after.

## Verification
- `yarn test`: 1131/1131 passed
- `npx tsc --noEmit`: clean
- Python: `pytest` 97/97 passed, `ruff check` clean, `mypy` clean

## Flagged for developer judgment (not fixed here)
See PR comment below — items needing a design/product decision, an SDK-surface change, or that couldn't be pinned to a confirmed advisory.

---
_Generated by [Claude Code](https://claude.ai/code/session_01APLGs6UxvVhopjC1x21VY2)_

### DIFF
--- a/sdk/python/src/shrtnr/client.py
+++ b/sdk/python/src/shrtnr/client.py
@@ -38,7 +38,9 @@ class Shrtnr:
         self._base_url = base_url.rstrip("/")
         self._api_key = api_key
         self._owns_client = http_client is None
-        self._http = http_client if http_client is not None else httpx.Client(timeout=timeout)
+        self._http = (
+            http_client if http_client is not None else httpx.Client(timeout=timeout, follow_redirects=True)
+        )
 
         self.links = Links(self._base_url, self._api_key, self._http)
         self.slugs = Slugs(self._base_url, self._api_key, self._http)
@@ -85,7 +87,9 @@ class AsyncShrtnr:
         self._base_url = base_url.rstrip("/")
         self._api_key = api_key
         self._owns_client = http_client is None
-        self._http = http_client if http_client is not None else httpx.AsyncClient(timeout=timeout)
+        self._http = (
+            http_client if http_client is not None else httpx.AsyncClient(timeout=timeout, follow_redirects=True)
+        )
 
         self.links = AsyncLinks(self._base_url, self._api_key, self._http)
         self.slugs = AsyncSlugs(self._base_url, self._api_key, self._http)
--- a/sdk/python/src/shrtnr/client.py
+++ b/sdk/python/src/shrtnr/client.py
@@ -39,7 +39,9 @@ class Shrtnr:
         self._api_key = api_key
         self._owns_client = http_client is None
         self._http = (
-            http_client if http_client is not None else httpx.Client(timeout=timeout, follow_redirects=True)
+            http_client
+            if http_client is not None
+            else httpx.Client(timeout=timeout, follow_redirects=True)
         )
 
         self.links = Links(self._base_url, self._api_key, self._http)
@@ -88,7 +90,9 @@ class AsyncShrtnr:
         self._api_key = api_key
         self._owns_client = http_client is None
         self._http = (
-            http_client if http_client is not None else httpx.AsyncClient(timeout=timeout, follow_redirects=True)
+            http_client
+            if http_client is not None
+            else httpx.AsyncClient(timeout=timeout, follow_redirects=True)
         )
 
         self.links = AsyncLinks(self._base_url, self._api_key, self._http)
--- a/sdk/python/tests/test_async_client.py
+++ b/sdk/python/tests/test_async_client.py
@@ -49,6 +49,18 @@ async def test_async_auth_sends_bearer_header(client: AsyncShrtnr) -> None:
     assert route.calls[0].request.headers["Authorization"] == f"Bearer {API_KEY}"
 
 
+@respx.mock
+async def test_async_follows_redirects_on_the_owned_client(client: AsyncShrtnr) -> None:
+    # See test_follows_redirects_on_the_owned_client in test_client.py:
+    # httpx.AsyncClient defaults to follow_redirects=False, unlike the
+    # TS/Dart SDKs, so the owned client must opt in explicitly.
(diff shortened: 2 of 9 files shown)


DennisAlund:
Items found during this review that need a design/product call, touch a public surface, or couldn't be pinned to a confirmed advisory. Not included in the diff.

**API / MCP**
- `create_link` via the MCP tool (`src/mcp/server.ts`) never triggers `autoLabelLink`/`fetchPageTitle` when no label is given, unlike the admin and REST paths (`src/index.tsx`, `src/api/links.ts`). Links created through MCP with no explicit label permanently lack the auto-fetched title. Fixing it needs a decision on mechanism: `McpAgent` runs as a Durable Object handler, and whether a `waitUntil`-equivalent is available/appropriate there (vs. blocking the tool response on an outbound fetch) isn't a pure local fix.

**Data layer**
- D1's 100-bound-parameter cap is unresolved for bundle *listing* (as opposed to *listing members of one bundle*, already fixed twice this cycle). `BundleManagementService.listBundles` (`src/services/bundle-management.ts:93`) and five queries in `ClickRepository.getBundleSummariesBulk` (`src/db/click-repository.ts`, roughly lines 1445–1555) all bind `...bundleIds`. An account with enough bundles (lower with an active range filter, since two of the queries bind extra params) will 500 the admin bundle overview page and the MCP `list_bundles` tool. Same fix technique as the prior two commits (resolve membership via subquery/join instead of binding an id array), but it touches 6 query sites across 2 files and needs a decision on how to carry the same archived/createdBy filter into each downstream query — a small design task, not a one-liner.

**Dependency / version drift** (no drop-in-safe security patch identified; nothing applied)
- `src/api/schemas.ts` and `src/mcp/server.ts` use Zod v4's deprecated chained `.url()`/`.email()` validators instead of the top-level `z.url()`/`z.email()`. Still supported as compat aliases, not broken, but worth a deliberate migration since it touches OpenAPI request schemas and MCP tool input schemas.
- `@modelcontextprotocol/sdk` is exact-pinned (no caret) in root `package.json`, coupled to `agents`' own exact pin — blocks even a security patch until `agents` moves. Needs manual verification against that SDK's advisories; none confirmed from this review.
- `agents` (`^0.19.0`) and `@cloudflare/workers-oauth-provider` (`^0.8.2`) are both pre-1.0. The eventual `agents` 0.20 migration (split MCP client/server v2 packages) is real future work, already tracked.
- `wrangler` (`^4.114.0`, devDependency) and Python's `httpx` (`>=0.28.1`) currency couldn't be checked against upstream from this environment — flagging for manual `npm view`/PyPI check rather than asserting a version.
- `@cloudflare/workers-types` is pinned `^5.20260728.1`, dated after the declared `compatibility_date: "2026-04-01"` in `wrangler.jsonc`. No concrete breakage found (only long-stable APIs like `caches.default`/`crypto.randomUUID` are in use), but worth a sanity check.

**Cross-SDK, low confidence**
- Neither the TypeScript SDK (bare `fetch`) nor the Dart SDK (`http.Client()`) sets a default request timeout, unlike Python (`httpx` with `DEFAULT_TIMEOUT`). A server that accepts the connection but never responds hangs the calling code indefinitely. A workaround exists today (inject a pre-configured client), so this is a behavior-parity question rather than a hard failure — flagging since a fix would mean a coordinated change across two SDKs' public construction options.
- All three SDKs wrap body-read failures on the *success* path in `ShrtnrError(0, ...)`, but not on the *error* (non-2xx) path — a mid-stream failure while reading an error response body reports the HTTP status rather than status 0. This is identical across TS, Python, and Dart, so it's consistent behavior rather than a single-SDK bug; noting it since it's a real inconsistency in how transport failures surface, but changing it is a judgment call on all three SDKs at once.


---

PR #41, opened 2026-08-22
Fix links pagination overflow for large datasets

imajus:
## Problem

The links pagination breaks the layout with a large number of links.

Screenshot From 2026-08-22 21-03-37

## Summary

- Render a compact seven-item pagination window with ellipses instead of one button per page.
- Preserve first, current, nearby, previous, next, and last-page navigation.
- Wrap the paginator and keep its controls contained at mobile widths.
- Mark the active page with aria-current and cover the pagination window with unit tests.

## Testing

- Targeted pagination and links-page tests: 17 passed.
- TypeScript check: passed.
- Playwright verification with 4,700 links at desktop and 375 px mobile widths: no horizontal page or control overflow.

### DIFF
--- a/src/__tests__/unit/links-pagination.test.ts
+++ b/src/__tests__/unit/links-pagination.test.ts
@@ -1,0 +1,39 @@
+// Copyright 2026 Oddbit (https://oddbit.id)
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, expect, it } from "vitest";
+import { paginationItems } from "../../pages/links";
+
+describe("links pagination window", () => {
+  it("shows every page when the result fits in the compact window", () => {
+    expect(paginationItems(1, 7)).toEqual([1, 2, 3, 4, 5, 6, 7]);
+  });
+
+  it("keeps the first pages and last page visible near the start", () => {
+    expect(paginationItems(1, 188)).toEqual([1, 2, 3, 4, 5, "ellipsis", 188]);
+  });
+
+  it("centers the current page between stable first and last links", () => {
+    expect(paginationItems(94, 188)).toEqual([
+      1,
+      "ellipsis",
+      93,
+      94,
+      95,
+      "ellipsis",
+      188,
+    ]);
+  });
+
+  it("keeps the first page and last pages visible near the end", () => {
+    expect(paginationItems(188, 188)).toEqual([
+      1,
+      "ellipsis",
+      184,
+      185,
+      186,
+      187,
+      188,
+    ]);
+  });
+});
--- a/src/pages/links.tsx
+++ b/src/pages/links.tsx
@@ -19,6 +19,43 @@ function formatDate(ts: number, lang: string): string {
 
 export type LinksFilter = "active" | "disabled" | "all";
 
+export type PaginationItem = number | "ellipsis";
+
+export function paginationItems(
+  currentPage: number,
+  totalPages: number,
+): PaginationItem[] {
+  if (totalPages <= 7) {
+    return Array.from({ length: totalPages }, (_, index) => index + 1);
+  }
+
+  if (currentPage <= 4) {
+    return [1, 2, 3, 4, 5, "ellipsis", totalPages];
+  }
+
+  if (currentPage >= totalPages - 3) {
+    return [
+      1,
+      "ellipsis",
+      totalPages - 4,
+      totalPages - 3,
+      totalPages - 2,
+      totalPages - 1,
+      totalPages,
+    ];
+  }
+
+  return [
+    1,
+    "ellipsis",
+    currentPage - 1,
+    currentPage,
+    currentPage + 1,
+    "ellipsis",
+    totalPages,
+  ];
+}
+
 type Props = {
   links: LinkWithSlugs[];
   sort: string;
@@ -274,13 +311,16 @@ export const LinksPage: FC<Props> = ({
                 >
                   <span class="icon icon-sm">chevron_left</span>
                 </a>
-                {Array.from({ length: totalPages }, (_, i) => i + 1).map(
-                  (p) => (
+                {paginationItems(currentPage, totalPages).map((item) =>
+                  item === "ellipsis" ? (
+                    <span class="page-ellipsis" aria-hidden="true">…</span>
+                  ) : (
                     <a
-                      class={`page-btn${p === currentPage ? " active" : ""}`}
-                      href={pageUrl(p)}
+                      class={`page-btn${item === currentPage ? " active" : ""}`}
+                      href={pageUrl(item)}
+                      aria-current={item === currentPage ? "page" : undefined}
                     >
-                      {p}
+                      {item}
                     </a>
                   ),
                 )}
--- a/src/styles.ts
+++ b/src/styles.ts
@@ -476,13 +476,14 @@ select.form-input { appearance: none; -webkit-appearance: none; padding-right: 2
 .link-disabled { opacity: 0.5; }
 .disabled-badge { display: inline-flex; align-items: center; gap: 0.2rem; background: var(--color-danger); color: var(--color-danger-foreground); font-size: 0.65rem; font-weight: 700; padding: 0.15rem 0.5rem; border-radius: var(--radius-md); text-transform: uppercase; letter-spacing: 0.05em; }
 .link-date { font-size: 0.7rem; color: var(--color-text-muted); margin-top: 0.2rem; }
-.pagination { display: flex; align-items: center; justify-content: space-between; gap: 1rem; margin-top: 1.4rem; padding: 0.75rem 0; }
+.pagination { display: flex; align-items: center; justify-content: space-between; flex-wrap: wrap; gap: 1rem; margin-top: 1.4rem; padding: 0.75rem 0; }
 .pagination-summary { font-size: 0.78rem; color: var(--color-text-muted); font-variant-numeric: tabular-nums; }
-.pagination-pages { display: flex; gap: 0.25rem; }
+.pagination-pages { display: flex; align-items: center; gap: 0.25rem; max-width: 100%; }
 .page-btn { background: transparent; border: 2px solid var(--color-border); border-radius: var(--radius-md); color: var(--color-text-muted); font-family: var(--font-family-body); font-size: 0.8rem; font-weight: 600; padding: 0.3rem 0.6rem; cursor: pointer; transition: all 0.2s; min-width: 2rem; text-align: center; text-decoration: none; display: inline-flex; align-items: center; justify-content: center; }
 .page-btn:hover { border-color: var(--color-text-muted); color: var(--color-text); }
 .page-btn.active { border-color: var(--color-accent); color: var(--color-accent); }
 .page-btn:disabled, .page-btn.disabled { opacity: 0.3; cursor: default; pointer-events: none; }
+.page-ellipsis { display: inline-flex; align-items: center; justify-content: center; min-width: 1.5rem; color: var(--color-text-subtle); font-size: 0.8rem; }
 .per-page { display: flex; align-items: center; gap: 0.5rem; font-size: 0.78rem; color: var(--color-text-muted); }
 .per-page-label { font-size: 0.78rem; color: var(--color-text-muted); }
 .per-page-select { min-width: 4rem; }
@@ -840,7 +841,9 @@ select.form-input { appearance: none; -webkit-appearance: none; padding-right: 2
   .filter-chip { flex: 1; justify-content: center; }
   .settings-layout { flex-direction: column !important; }
   .settings-layout > div:last-child { max-width: 100% !important; }
-  .pagination { flex-wrap: wrap; gap: 0.5rem; justify-content: center; }
+  .pagination { gap: 0.5rem; justify-content: center; }
+  .pagination-pages { order: 3; width: 100%; justify-content: center; overflow-x: auto; }
+  .page-btn { padding-inline: 0.45rem; }
   .bundle-grid { grid-template-columns: 1fr; }
   .detail-analytics { grid-template-columns: 1fr; }
   .bundle-link-head { gap: 0.5rem; }


DennisAlund (commented):
Reviewed the compact-window logic and it holds up. I traced the branch boundaries: no case produces a single-page ellipsis, because the `<= 7` early return absorbs every total the three branches would otherwise mishandle, and the output is always exactly seven items. `page-btn` markup lives in one place, so this covers every paginated view in the admin UI. The tests add without touching existing ones, which matches our testing rule.

Five things below. The first three break the mobile case this PR sets out to fix, so I'd land them before merge.

One judgment call I am not asking you to change: the literal `…` in JSX sits against our i18n rule, but it is an `aria-hidden` decorative glyph rather than copy. Leaving it is fine by me.

The review also surfaced problems that predate this PR and are ours, not yours: the prev/next chevrons have no accessible name (`linkDetail.prevPage` and `linkDetail.nextPage` already exist in the locale files and go unused), the disabled chevrons stay keyboard-focusable with `href="#"`, the paginator renders for single-page results because its guard tests the unfiltered count against a hardcoded `25`, and the page still loads and delta-enriches every row per request with no SQL LIMIT/OFFSET. That last one is the real cost of the 4,700-link case: the paginator stops overflowing, but the next symptom is CPU limits. Filed as separate issues.

Heads-up on overlap: #42 touches `buildUrl` in this same file to preserve the `search` param across pagination. Different region, so I don't expect a textual conflict, but whichever lands second is worth a re-run.

src/styles.ts
```diff
@@ -840,7 +841,9 @@ select.form-input { appearance: none; -webkit-appearance: none; padding-right: 2
   .filter-chip { flex: 1; justify-content: center; }
   .settings-layout { flex-direction: column !important; }
   .settings-layout > div:last-child { max-width: 100% !important; }
-  .pagination { flex-wrap: wrap; gap: 0.5rem; justify-content: center; }
+  .pagination { gap: 0.5rem; justify-content: center; }
+  .pagination-pages { order: 3; width: 100%; justify-content: center; overflow-x: auto; }
```
DennisAlund, line 845: `justify-content: center` and `overflow-x: auto` on the same element leaves the start edge unreachable. A centered flex container that overflows pushes content past its start edge, and browsers expose no scroll range in that direction, so the `1` button and the first ellipsis clip off-screen with no way to scroll or tab back to them. Reproduces at 320px, and at wider widths once text zoom reaches 150%.

Set `justify-content: flex-start` on the scroll container and center the content with `margin: auto` on the children, or use `justify-content: safe center`.

DennisAlund, line 845: `order: 3` desyncs focus order from visual order. DOM order inside `.pagination` is summary, pages, per-page. With `order: 3` the mobile layout puts summary and per-page on the first row and the page links on the second, but tab order follows the DOM, so focus jumps from the page links back up to the per-page `<select>` rendered above them. That is a WCAG 2.4.3 focus-order mismatch, and a sighted keyboard user watches the focus ring travel upward against the reading direction.

Reordering the JSX children in `LinksPage` and dropping `order` produces the same layout with matching focus order.

src/styles.ts
```diff
@@ -840,7 +841,9 @@ select.form-input { appearance: none; -webkit-appearance: none; padding-right: 2
...
+  .page-btn { padding-inline: 0.45rem; }
```
DennisAlund, line 846: `.page-btn` keeps the default `flex-shrink: 1`, so the `overflow-x: auto` added on the line above never engages. Flexbox shrinks items before a container overflows: at 360px with three-digit page numbers each button squeezes to the `min-width: 2rem` floor from line 482 while `padding-inline` and the 2px borders stay fixed, so the digits render past their own borders instead of the row scrolling.

Add `flex-shrink: 0` here.

src/__tests__/unit/links-pagination.test.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright 2026 Oddbit (https://oddbit.id)
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, expect, it } from "vitest";
+import { paginationItems } from "../../pages/links";
+
+describe("links pagination window", () => {
```
DennisAlund, line 7: These cover the pure array function well, but nothing asserts the rendered markup, so two behaviors this PR introduces go unguarded: `aria-current="page"` on the active link, and the `page-ellipsis` span. Dropping `aria-current` or regressing the ellipsis back to a `page-btn` would keep the suite green.

`src/__tests__/page/links-page.test.ts` already fetches the page and regex-matches HTML, so it is the natural home for both assertions.

Worth one more case here too: `totalPages === 8`. That is the boundary where the middle branch is unreachable and `currentPage === 5` has to fall through to the near-end branch.

src/pages/links.tsx
```diff
@@ -19,6 +19,43 @@ function formatDate(ts: number, lang: string): string {
 
 export type LinksFilter = "active" | "disabled" | "all";
 
+export type PaginationItem = number | "ellipsis";
+
+export function paginationItems(
+  currentPage: number,
+  totalPages: number,
+): PaginationItem[] {
+  if (totalPages <= 7) {
```
DennisAlund, line 28: `paginationItems` is exported with no guard on its inputs: `paginationItems(1, 0)` returns `[]`, which renders a paginator with two chevrons and zero page links. The one current call site is safe only because of `Math.max(1, ...)` on `totalPages` and the upstream page clamp in `src/index.tsx`, and neither guarantee lives in this function.

We are migrating the admin UI to htmx widget islands, so the next caller is likely a partial that will not repeat both clamps. Clamping here, with `totalPages = Math.max(1, totalPages)` and `currentPage` pinned into `[1, totalPages]`, makes the contract self-contained.

DennisAlund:
Filed the pre-existing items from my review so they don't muddy this PR:

- #43 paginator controls are inaccessible to keyboard and screen readers (unnamed chevrons, focusable disabled links, no landmark)
- #44 paginator renders for single-page results
- #45 links page loads and delta-enriches the whole catalog per request

The `search` param drop is already handled in #42, so nothing to do there.

None of those are yours to fix in this PR. The five inline comments are the ones I'd like addressed here, and the first three are the merge blockers.

src/styles.ts
DennisAlund: Reproduced in Chromium against the real `adminStyles`, with the paginator constrained to the admin column width and three-digit page numbers. With `justify-content: center` the row overflows past its start edge while `scrollLeft` sits at its minimum, so the clipped part has no scroll range to reach it:

| viewport | text zoom | hidden past the start edge |
| --- | --- | --- |
| 320px | 100% | 8px |
| 320px | 150% | 28px |
| 375px | 150% | 1px |

Took the auto-margin option rather than `safe center`, so nothing depends on `safe` keyword support. Auto margins absorb positive free space and resolve to 0 against negative free space, which centers the window while it fits and start-aligns it once it overflows.

Measured after the change: 0px hidden at 320, 360 and 375px, at both zoom levels. Centering still holds where the row fits, with lead gap equal to trail gap to the pixel at 700px and 768px.

The replacement drops `order: 3` as well, since both faults live on this one line. Details on that in the sibling thread.

```suggestion
  /* The flex line follows DOM order, with no order property, so focus travels with
     the reading direction. Auto margins center the window while it fits and collapse
     to 0 once it overflows, keeping the scroll container's start edge reachable. */
  .pagination-pages { width: 100%; overflow-x: auto; }
  .pagination-pages > :first-child { margin-inline-start: auto; }
  .pagination-pages > :last-child { margin-inline-end: auto; }
```

DennisAlund: The mismatch reproduces. Reading the `top` of every focusable in tab order at 360px, before the change: `[510, 512, 512, 512, 512, 512, 510, 180]`. The `<select>` sits 330px above the page links that precede it in the DOM, so focus travels back up the page. After dropping `order`, the same trace rises monotonically.

One correction to the suggested fix: reordering the JSX is not needed, and it would cost more than it fixes.

Screenshots at 360px show `.pagination` already breaking into three rows either way, because `.pagination-pages` is `width: 100%` and cannot share a line with anything. `order: 3` was not packing summary and per-page onto one row; it was swapping rows 2 and 3. Dropping it alone restores summary, pages, per-page at the same three-row height, now with focus following the visual order.

Reordering the children would also reach desktop, where `.pagination` is `justify-content: space-between` and DOM order alone produces `summary | pages | per-page`. Moving per-page ahead of pages in the JSX would rearrange that, and restoring it would need `order` on desktop, which relocates the same WCAG problem rather than removing it. Leaving the JSX alone keeps every change inside the `max-width: 768px` block, so desktop CSS is byte-identical.

Fix rides in the suggestion on the sibling thread, since both faults sit on line 845.

src/styles.ts
DennisAlund: Confirmed, and the measurement matches the reasoning. Element widths at 360px before the change, against natural widths after:

| item | before | after |
| --- | --- | --- |
| chevron | 32.9px | 36.4px |
| `187` | 36.5px | 40.8px |
| `999` | 39.0px | 43.9px |

The row compressed toward the `min-width` floor instead of overflowing, so `.pagination-pages` reported `scrollWidth - clientWidth` of 0 at 360px and 375px and `overflow-x: auto` never engaged. With `flex-shrink: 0` the same row reports 25px and 10px of scrollable overflow and the buttons hold their natural size.

`.page-ellipsis` is in the suggestion too, though as intent rather than a fix: it measured 24.0px before and after, because its `min-width: 1.5rem` already equals its content width and floors any shrink. It would start shrinking the moment that rule gains padding. Drop that line if you would rather not carry it.

```suggestion
  .page-btn { padding-inline: 0.45rem; flex-shrink: 0; }
  .page-ellipsis { flex-shrink: 0; }
```

src/__tests__/unit/links-pagination.test.ts
DennisAlund: Both gaps are real. I checked by mutation rather than by reading: deleting the `aria-current` attribute, and regressing the ellipsis `<span>` back to `class="page-btn"`, each left the suite green before these assertions and each fails against them now.

The suggestion below adds the `totalPages === 8` boundary. Page 5 is both greater than 4 and at least `totalPages - 3`, so it falls through to the near-end window and the centered branch is unreachable at that total. Page 4 is included as the last total that still takes the near-start branch. The two clamping cases belong to the `paginationItems` thread, but they live in this file, so they ride along here to keep a single suggestion per file.

```suggestion
describe("links pagination window", () => {
  it("falls through to the near-end window at the 8-page boundary", () => {
    // 8 is the smallest total the compact window applies to, and the one total
    // where the centered branch is unreachable: page 5 is both > 4 and >= 8 - 3.
    expect(paginationItems(5, 8)).toEqual([1, "ellipsis", 4, 5, 6, 7, 8]);
    expect(paginationItems(4, 8)).toEqual([1, 2, 3, 4, 5, "ellipsis", 8]);
  });

  it("clamps a total below one page to a single-page window", () => {
    expect(paginationItems(1, 0)).toEqual([1]);
  });

  it("clamps a current page outside the total into range", () => {
    expect(paginationItems(0, 5)).toEqual([1, 2, 3, 4, 5]);
    expect(paginationItems(400, 188)).toEqual([
      1,
      "ellipsis",
      184,
      185,
      186,
      187,
      188,
    ]);
  });

```

The markup assertions cannot be a suggestion: `src/__tests__/page/links-page.test.ts` is not among this PR's three changed files, so GitHub has no diff line to anchor one to. Adding this case to that file covers both. `per_page=1` is the cheap way past the 7-item window, since the param clamps at 1 rather than to the 25/50/100 options, so 8 links give 8 pages:

```ts
  it("marks the active page with aria-current and renders ellipses outside the window", async () => {
    // per_page=1 gives 8 pages from 8 links, past the 7-item compact window.
    for (let i = 0; i < 8; i++) {
      await LinkRepository.create(env.DB, {
        url: `https://example${i}.com`,
        slug: `s${i}`,
      });
    }
    const res = await SELF.fetch(req("/_/admin/links?per_page=1&page=1"));
    const html = await res.text();
    // Page 1 is the active link and carries aria-current.
    expect(html).toMatch(/<a[^>]*class="page-btn active"[^>]*aria-current="page"[^>]*>1</);
    // Pages 6 and 7 are outside the window, replaced by an ellipsis span.
    expect(html).toMatch(/<span class="page-ellipsis"/);
    expect(html).not.toMatch(/class="page-btn[^"]*"[^>]*>6</);
  });
```

With all five items applied: 1136 tests pass across 77 files, and `tsc --noEmit` is clean.

src/pages/links.tsx
DennisAlund: `paginationItems(1, 0)` returns `[]` as described, and the two guards that make the current call site safe do live elsewhere. Clamping inside the function.

Reassigning the parameters rather than introducing locals, which is the form your comment used, keeps the rest of the body untouched and makes this a five-line insertion instead of a rewrite of all four branches.

```suggestion
  // Clamp here rather than at the call site: no caller should have to repeat
  // these guards, and the next one will be an htmx partial that does not.
  totalPages = Math.max(1, totalPages);
  currentPage = Math.min(Math.max(1, currentPage), totalPages);

  if (totalPages <= 7) {
```

Tests for this ride in the suggestion on `links-pagination.test.ts`, since they share that file with the boundary case: `paginationItems(1, 0)` returns `[1]`, and a `currentPage` outside `[1, totalPages]` pins to the nearest end.

imajus:
@DennisAlund applied all suggested changes.

DennisAlund (approved):
Thanks

src/styles.ts
DennisAlund: Applied in #47. Measured after the change: 0px hidden past the start edge at 320, 360 and 375px, at both rem scales, against 8px and 92px before. Centering still holds where the row fits, lead equal to trail to the pixel at 700px and 768px.

DennisAlund: Applied in #47, by dropping `order` alone. The mismatch reproduced: the `<select>` renders 40px above the page links that precede it in the DOM (73 vs 113), and sits below them after. The JSX is untouched, so desktop CSS stays byte-identical — every changed line is inside the `max-width: 768px` block, which opens at line 683 and closes at 760.

src/styles.ts
DennisAlund: Applied in #47. The `188` button pins to its 32px `min-width` floor before the change and holds 41.8px natural width after, so the row overflows and scrolls instead of compressing. `.page-ellipsis` carried along as intent: it measured identically before and after, since its `min-width: 1.5rem` already floors any shrink.

src/__tests__/unit/links-pagination.test.ts
DennisAlund: Applied in #47, including the `totalPages === 8` boundary. Both markup gaps confirmed by mutation rather than by reading: deleting `aria-current` and regressing the ellipsis `<span>` back to `class="page-btn"` each leave the suite green today and each fail against the new assertions.

src/pages/links.tsx
DennisAlund: Applied in #47, clamping inside the function by reassigning the parameters. Verified by mutation: removing the two clamp lines fails the new `paginationItems(1, 0)` case.

DennisAlund:
Merged as-is; all five review items land in #47 against `main`, verified at 1149 tests across 77 files with `tsc --noEmit` clean. Items 1–3 were measured in Chromium against the real extracted `adminStyles` rather than reasoned about; items 4–5 by mutation. Per-item notes are in their threads above.

PR #42, opened 2026-08-22
Fix search pagination state

imajus:
## Problem

Links search query is reset when navigated to other pages.

## Summary

- preserve the active search query in links pagination, sorting, filtering, and page-size URLs
- add a regression test for navigating to page 2 of search results

## Testing

- `npm test -- --run src/__tests__/page/links-page.test.ts`

### DIFF
--- a/src/__tests__/page/links-page.test.ts
+++ b/src/__tests__/page/links-page.test.ts
@@ -174,6 +174,26 @@ describe("Links listing page", () => {
     expect(html).toMatch(/1\s*[–-]\s*25\s+of\s+30/);
   });
 
+  it("preserves the active search in pagination URLs", async () => {
+    for (let i = 0; i < 30; i++) {
+      await LinkRepository.create(env.DB, {
+        url: `https://pdf.co/document-${i}`,
+        slug: `pdf-${i}`,
+      });
+    }
+
+    const res = await SELF.fetch(req("/_/admin/links?search=pdf.co"));
+    const html = await res.text();
+    const pageTwoHref = html.match(
+      /class="page-btn[^"]*" href="([^"]+)"[^>]*>\s*2\s*<\/a>/,
+    )?.[1];
+    const decodedHref = pageTwoHref?.replaceAll("&amp;", "&");
+
+    expect(pageTwoHref).toBeDefined();
+    expect(decodedHref).toContain("page=2");
+    expect(decodedHref).toContain("search=pdf.co");
+  });
+
   it("clamps a negative page param instead of producing an empty, out-of-range slice", async () => {
     for (let i = 0; i < 30; i++) {
       await LinkRepository.create(env.DB, {
--- a/src/__tests__/page/links-page.test.ts
+++ b/src/__tests__/page/links-page.test.ts
@@ -194,6 +194,88 @@ describe("Links listing page", () => {
     expect(decodedHref).toContain("search=pdf.co");
   });
 
+  // Pull hrefs out of anchors carrying `className`, independent of attribute
+  // order, with entity-encoded ampersands decoded for plain substring checks.
+  // `label` matches the anchor's visible text once tags are stripped.
+  function anchorHrefs(html: string, className: string, label?: string): string[] {
+    const found: string[] = [];
+    const anchor = /<a\b([^>]*)>([\s\S]*?)<\/a>/g;
+    let m: RegExpExecArray | null;
+    while ((m = anchor.exec(html)) !== null) {
+      const [, attrs, inner] = m;
+      if (!new RegExp(`class="[^"]*\\b${className}\\b[^"]*"`).test(attrs)) continue;
+      if (label !== undefined && inner.replace(/<[^>]*>/g, "").trim() !== label) continue;
+      const href = attrs.match(/href="([^"]*)"/)?.[1];
+      if (href) found.push(href.replaceAll("&amp;", "&"));
+    }
+    return found;
+  }
+
+  async function seedSearchFixture(): Promise<void> {
+    for (let i = 0; i < 30; i++) {
+      await LinkRepository.create(env.DB, {
+        url: `https://pdf.co/document-${i}`,
+        slug: `pdf-${i}`,
+      });
+    }
+    for (let i = 0; i < 10; i++) {
+      await LinkRepository.create(env.DB, {
+        url: `https://other.example/note-${i}`,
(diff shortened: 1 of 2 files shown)


DennisAlund:
Thank you for finding this and for the contribution.

src/__tests__/page/links-page.test.ts
```diff
@@ -174,6 +174,26 @@ describe("Links listing page", () => {
...
+    const html = await res.text();
+    const pageTwoHref = html.match(
+      /class="page-btn[^"]*" href="([^"]+)"[^>]*>\s*2\s*<\/a>/,
+    )?.[1];
+    const decodedHref = pageTwoHref?.replaceAll("&amp;", "&");
+
+    expect(pageTwoHref).toBeDefined();
+    expect(decodedHref).toContain("page=2");
+    expect(decodedHref).toContain("search=pdf.co");
+  });
```
DennisAlund, line 195: This appends to your test rather than editing it, so your assertions stay exactly as written.

Three additions:

- **Follows the page-two href.** Yours proves the link is built right; this proves the server acts on it. Seeds 10 non-matching links alongside the 30, then fetches the href and checks the summary reads `26–30 of 30`, no unmatched link leaks in, and the search box stays populated.
- **Covers the other three URL builders** your description mentions: sort, filter, and page size each keep the query and reset to page one.
- **A negative case**, so the param disappears once the query is cleared and we do not just assert one direction.

`anchorHrefs` reads hrefs independent of attribute order, so these survive markup churn rather than pinning to today's attribute sequence.

Verified on your branch: 17 passing, `tsc --noEmit` clean. Both new positive cases fail if the `search: searchQuery` line is removed, so they guard your fix rather than restating it.

```suggestion
  });

  // Pull hrefs out of anchors carrying `className`, independent of attribute
  // order, with entity-encoded ampersands decoded for plain substring checks.
  // `label` matches the anchor's visible text once tags are stripped.
  function anchorHrefs(html: string, className: string, label?: string): string[] {
    const found: string[] = [];
    const anchor = /<a\b([^>]*)>([\s\S]*?)<\/a>/g;
    let m: RegExpExecArray | null;
    while ((m = anchor.exec(html)) !== null) {
      const [, attrs, inner] = m;
      if (!new RegExp(`class="[^"]*\\b${className}\\b[^"]*"`).test(attrs)) continue;
      if (label !== undefined && inner.replace(/<[^>]*>/g, "").trim() !== label) continue;
      const href = attrs.match(/href="([^"]*)"/)?.[1];
      if (href) found.push(href.replaceAll("&amp;", "&"));
    }
    return found;
  }

  async function seedSearchFixture(): Promise<void> {
    for (let i = 0; i < 30; i++) {
      await LinkRepository.create(env.DB, {
        url: `https://pdf.co/document-${i}`,
        slug: `pdf-${i}`,
      });
    }
    for (let i = 0; i < 10; i++) {
      await LinkRepository.create(env.DB, {
        url: `https://other.example/note-${i}`,
        slug: `other-${i}`,
      });
    }
  }

  it("lists only matching links when page two of a search is opened", async () => {
    await seedSearchFixture();

    const first = await SELF.fetch(req("/_/admin/links?search=pdf.co"));
    const [pageTwo] = anchorHrefs(await first.text(), "page-btn", "2");
    expect(pageTwo).toBeDefined();

    const res = await SELF.fetch(req(pageTwo));
    expect(res.status).toBe(200);
    const html = await res.text();

    // 30 matches, not the 40 links in the table.
    expect(html).toMatch(/26\s*[–-]\s*30\s+of\s+30/);
    expect(html).not.toContain("other.example");
    // The search box stays populated so the query is still editable.
    expect(html).toMatch(/id="quick-url"[^>]*value="pdf\.co"/);
  });

  it("carries the active search through sort, filter, and page-size URLs", async () => {
    await seedSearchFixture();

    const res = await SELF.fetch(req("/_/admin/links?search=pdf.co"));
    const html = await res.text();
    const perPageHrefs = [...html.matchAll(/<option value="([^"]*)"/g)].map((m) =>
      m[1].replaceAll("&amp;", "&"),
    );
    const controls = [
      ...anchorHrefs(html, "sort-btn"),
      ...anchorHrefs(html, "filter-chip"),
      ...perPageHrefs,
    ];

    expect(controls).toHaveLength(8);
    // Each control resets to page one but must keep the query.
    for (const href of controls) {
      expect(href).toContain("search=pdf.co");
      expect(href).toContain("page=1");
    }
  });

  it("drops the search from pagination URLs once the query is cleared", async () => {
    await seedSearchFixture();

    const res = await SELF.fetch(req("/_/admin/links"));
    const [pageTwo] = anchorHrefs(await res.text(), "page-btn", "2");

    expect(pageTwo).toBeDefined();
    expect(pageTwo).not.toContain("search=");
  });
```

---

DennisAlund (commented):
Thank you for finding this and for the contribution. I validated the fix end to end before suggesting anything, and it holds up.

**The one-line change is the right one.** `sortUrl`, `pageUrl`, `perPageUrl`, and `filterUrl` all route through `buildUrl`, so adding `search` there covers every navigation control in a single place rather than patching four call sites. Nice spot.

**Your regression test is a real one.** I deleted the `search: searchQuery` line and watched it fail, restored it and watched it pass. Full suite green at 1129 tests across 76 files, `tsc --noEmit` clean, and no OpenAPI spec drift, so nothing for the SDK hashes to chase.

I also checked the round trip rather than the href alone: page two of a search returns `26–30 of 30` with no unmatched links leaking in, and the search box stays populated. The bug is genuinely fixed, not just the link.

**One suggestion inline**, appending to your test rather than rewriting it. It follows the page-two href to prove the server scopes the results, covers the sort, filter, and page-size URLs your description also mentions, and adds a negative case for the cleared-query direction. Commit it or skip it, whichever you prefer.

**A separate finding, not for this PR.** Pagination is one way to lose a search; opening a result is the other. `src/pages/link-detail.tsx` hardcodes its back link to a bare `/_/admin/links`, so searching, opening a result, and hitting back drops you on page one of the unfiltered list. Fixing it touches `link-detail.tsx` and `index.tsx`, both outside this diff, so it cannot ride along as a suggestion here. I have a tested patch for it and am happy to open a follow-up PR, or leave it to you if you would rather take it. Either way it is not a blocker for this one.

One heads-up for the maintainers: CI has not run on this PR yet. Fork PRs sit at zero checks until someone approves the workflow run, so the numbers above are from local runs.

---

PR #48, opened 2026-08-26
Hide the links paginator when the result set fits one page

### LINKED ISSUE #44: Links paginator renders for single-page results
Surfaced while reviewing #41. Predates that PR.

`src/pages/links.tsx:261` guards the paginator with:

```jsx
{(totalPages > 1 || links.length > 25) && (
```

Two problems. The condition tests `links`, the unfiltered set, rather than `sorted`, the set that pagination actually walks. It also hardcodes `25` instead of reading `perPage`.

Reproductions:

- `per_page=100` with 30 links. `totalPages` is 1, but `30 > 25` renders a paginator whose entire window is the single button `1` with both chevrons disabled.
- `filter=disabled` over 200 links of which 2 have expired. `sorted.length` is 2 and `totalPages` is 1, yet `200 > 25` renders the same dead control.

Fix: `totalPages > 1 || sorted.length > perPage`, or just `totalPages > 1`, since `totalPages` already derives from `sorted.length` and `perPage` at `src/pages/links.tsx:64`.

DennisAlund:
Closes [#44](https://github.com/oddbit/shrtnr/issues/44).

## The bug

`src/pages/links.tsx` guarded the paginator with `totalPages > 1 || links.length > 25`. That condition tests `links`, the unfiltered set, instead of `sorted`, the set pagination actually walks. It also hardcodes `25` instead of reading `perPage`.

Both reproductions from the issue render a dead control: a lone `1` button with both chevrons disabled.

- `per_page=100` over 30 links. `totalPages` is 1, but `30 > 25` renders the paginator.
- `filter=disabled` over 200 links of which 2 have expired. `sorted.length` is 2 and `totalPages` is 1, yet `200 > 25` renders it.

## The fix

Split the gate. `.pagination` renders whenever the table renders. Only `.pagination-pages`, the chevrons and page numbers, depends on `totalPages > 1`.

An earlier revision of this PR gated the whole `.pagination` block on `totalPages > 1`. That removed the dead control, but it also hid the result summary and the 25/50/100 per-page selector, which stranded the user. `buildUrl` re-emits `per_page` into every sort link and filter chip (`src/pages/links.tsx:116`) and the range picker preserves it (`:145`). After picking 50 rows over 30 links, `totalPages` fell to 1 and no in-page control set it back to 25. `per_page` lives only in the URL (`src/index.tsx:235`), so the only way back was leaving the links page entirely.

Issue #44 asked for the dead control to go, not for the summary and the selector to go with it.

Also collapsed the page-summary ternary. `totalPages > 1` implies `sorted.length > perPage >= 1`, so the `sorted.length === 0` branch was already unreachable once the `links.length > 25` escape hatch went away.

## Tests

Four cases in `src/__tests__/page/links-page.test.ts`, one per reproduction plus two guards:

- `per_page=100` with 30 links renders 30 rows and no page buttons, and keeps the summary and selector.
- `filter=disabled` with 2 of 30 links expired renders `s0` and `s1` only, no page buttons, and keeps the selector.
- 30 links at the default page size still renders the page buttons, page `2` included.
- `per_page=50` over 30 links still offers the `per_page=25` route back.

Every case asserts a 200 status and counts rendered rows, so none can pass on an empty state or a 500.

Verified three ways:

- Three of the four fail against the whole-block gate: both reproduction cases plus the `per_page=50` route back. The multi-page case exercises `totalPages > 1`, where either gate shape renders the block, so it passes against both by design.
- The two reproduction cases also fail against `main`, so they still cover the original bug.
- Full suite: 1153 tests across 77 files, all passing. `tsc --noEmit` clean.

## Follow-ups

Reviewing this PR surfaced five pre-existing bugs in `src/pages/links.tsx` that it does not touch. Filed separately to keep this change to one concern: #51 (empty-state copy ignores the active filter), #52 (zero-match search reads as an empty catalog), #53 (disabled chevrons stay keyboard-activatable), #54 (chevrons have no accessible name, page group has no landmark).

### DIFF
--- a/src/__tests__/page/links-page.test.ts
+++ b/src/__tests__/page/links-page.test.ts
@@ -224,4 +224,47 @@ describe("Links listing page", () => {
     const html = await res.text();
     expect(html).toMatch(/1\s*[–-]\s*1\s+of\s+30/);
   });
+
+  it("hides the paginator when a wider per_page fits every link on one page", async () => {
+    for (let i = 0; i < 30; i++) {
+      await LinkRepository.create(env.DB, {
+        url: `https://example${i}.com`,
+        slug: `s${i}`,
+      });
+    }
+    const res = await SELF.fetch(req("/_/admin/links?per_page=100"));
+    const html = await res.text();
+    // All 30 rows fit on page 1, so there is nothing to page through.
+    expect(html).not.toContain('class="pagination"');
+    expect(html).not.toMatch(/class="page-btn/);
+  });
+
+  it("hides the paginator when a filter narrows the result set to one page", async () => {
+    const now = Math.floor(Date.now() / 1000);
+    for (let i = 0; i < 30; i++) {
+      await LinkRepository.create(env.DB, {
+        url: `https://example${i}.com`,
+        slug: `s${i}`,
+        expiresAt: i < 2 ? now - 3600 : null,
+      });
+    }
+    const res = await SELF.fetch(req("/_/admin/links?filter=disabled"));
+    const html = await res.text();
+    // Only 2 of the 30 links are expired; the unfiltered count must not decide this.
+    expect(html).not.toContain('class="pagination"');
+    expect(html).not.toMatch(/class="page-btn/);
+  });
+
+  it("still renders the paginator when the result set spans more than one page", async () => {
+    for (let i = 0; i < 30; i++) {
+      await LinkRepository.create(env.DB, {
+        url: `https://example${i}.com`,
+        slug: `s${i}`,
+      });
+    }
+    const res = await SELF.fetch(req("/_/admin/links"));
+    const html = await res.text();
+    expect(html).toContain('class="pagination"');
+    expect(html).toMatch(/class="page-btn[^"]*"[^>]*>2</);
+  });
 });
--- a/src/__tests__/page/links-page.test.ts
+++ b/src/__tests__/page/links-page.test.ts
@@ -225,7 +225,7 @@ describe("Links listing page", () => {
     expect(html).toMatch(/1\s*[–-]\s*1\s+of\s+30/);
   });
 
-  it("hides the paginator when a wider per_page fits every link on one page", async () => {
+  it("hides the page buttons when a wider per_page fits every link on one page", async () => {
     for (let i = 0; i < 30; i++) {
       await LinkRepository.create(env.DB, {
         url: `https://example${i}.com`,
@@ -233,13 +233,19 @@ describe("Links listing page", () => {
       });
     }
     const res = await SELF.fetch(req("/_/admin/links?per_page=100"));
+    expect(res.status).toBe(200);
     const html = await res.text();
-    // All 30 rows fit on page 1, so there is nothing to page through.
-    expect(html).not.toContain('class="pagination"');
+    // All 30 rows land on page 1, so there is nothing to page through.
+    expect(html.match(/class="col-short-chip-slug"/g)).toHaveLength(30);
+    expect(html).not.toContain('class="pagination-pages"');
     expect(html).not.toMatch(/class="page-btn/);
+    // The summary and the per-page selector survive: they are the only way back to 25.
+    expect(html).toContain('class="pagination"');
+    expect(html).toContain('class="pagination-summary"');
+    expect(html).toContain("per-page-select");
   });
 
-  it("hides the paginator when a filter narrows the result set to one page", async () => {
+  it("hides the page buttons when a filter narrows the result set to one page", async () => {
     const now = Math.floor(Date.now() / 1000);
     for (let i = 0; i < 30; i++) {
       await LinkRepository.create(env.DB, {
@@ -249,13 +255,20 @@ describe("Links listing page", () => {
       });
     }
     const res = await SELF.fetch(req("/_/admin/links?filter=disabled"));
+    expect(res.status).toBe(200);
     const html = await res.text();
-    // Only 2 of the 30 links are expired; the unfiltered count must not decide this.
-    expect(html).not.toContain('class="pagination"');
+    // Only s0 and s1 are expired; the unfiltered count must not decide this.
+    expect(html.match(/class="col-short-chip-slug"/g)).toHaveLength(2);
+    expect(html).toContain(">s0<");
+    expect(html).toContain(">s1<");
+    expect(html).not.toContain(">s2<");
+    expect(html).not.toContain('class="pagination-pages"');
     expect(html).not.toMatch(/class="page-btn/);
+    expect(html).toContain('class="pagination-summary"');
+    expect(html).toContain("per-page-select");
   });
 
-  it("still renders the paginator when the result set spans more than one page", async () => {
+  it("still renders the page buttons when the result set spans more than one page", async () => {
     for (let i = 0; i < 30; i++) {
       await LinkRepository.create(env.DB, {
         url: `https://example${i}.com`,
@@ -263,8 +276,27 @@ describe("Links listing page", () => {
       });
     }
     const res = await SELF.fetch(req("/_/admin/links"));
+    expect(res.status).toBe(200);
     const html = await res.text();
     expect(html).toContain('class="pagination"');
+    expect(html).toContain('class="pagination-pages"');
     expect(html).toMatch(/class="page-btn[^"]*"[^>]*>2</);
+    expect(html).toContain("per-page-select");
   });
+
+  it("keeps the per-page selector reachable after widening per_page", async () => {
+    for (let i = 0; i < 30; i++) {
+      await LinkRepository.create(env.DB, {
+        url: `https://example${i}.com`,
+        slug: `s${i}`,
+      });
+    }
+    const res = await SELF.fetch(req("/_/admin/links?per_page=50"));
+    expect(res.status).toBe(200);
+    const html = await res.text();
+    // Every in-page link re-emits per_page, so the selector is the only route back to 25.
+    expect(html).toContain("per-page-select");
+    expect(html).toContain("per_page=25");
+  });
+
 });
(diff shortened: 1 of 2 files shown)


DennisAlund (commented):
Reviewed the paginator gate. Ten findings, all verified against rendered output rather than by reading alone. Four sat inside this diff and are fixed in 741b24a; five are pre-existing bugs in code this PR never touches, filed separately; one I skipped.

### Fixed in 741b24a

Inline comments below cover these: the per-page selector trap, the dead page-summary ternary, and two gaps in the added tests.

The headline problem: gating the whole `.pagination` block on `totalPages > 1` hid the result summary and the 25/50/100 selector along with the dead page buttons. `buildUrl` re-emits `per_page` into every sort link, filter chip, and the range picker's preserved params, and `per_page` lives only in the URL (`src/index.tsx:235`). After picking 50 rows over 30 links, no in-page control set it back to 25.

### Pre-existing, filed rather than fixed here

These live outside the diff, so GitHub cannot anchor them inline. One coherent change per commit, so they stay out of this PR:

- `src/pages/links.tsx:222` picks empty-state copy from `links.length > 0`, ignoring the active filter. `filter=disabled` with nothing expired renders `All links are disabled. Toggle "Show disabled"` while `toolbar-count` reads `30 links`. The `links.showDisabled` key it points at is referenced nowhere outside the i18n files. → #51
- `src/pages/links.tsx:222` again: the route sets `links` to search results (`src/index.tsx:230`), so a zero-match search renders `No links yet` over a 200-link catalog. → #52
- `src/pages/links.tsx:313-318` and `:332-341` keep the disabled chevrons as anchors with `href="#"`. `.page-btn.disabled { pointer-events: none }` (`src/styles.ts:485`) blocks the mouse but not Enter, so tabbing to a disabled chevron and pressing Enter navigates to `#` and jumps the viewport. → #53
- The icon-only chevrons carry no accessible name, so screen readers announce `chevron_left, link`. Repo convention is `aria-label={t(...)}` (`src/pages/keys.tsx:92`, `src/pages/layout.tsx:117`). `.pagination-pages` also lacks the `role="group"` + `aria-label` that the filter chips (`:185`) and range picker (`src/components/range-picker.tsx:23`) both carry. → #54

### Skipped

The 30-link seeding loop is copy-pasted six times in `links-page.test.ts`. Accurate, but extracting a helper rewrites three pre-existing tests for no behavior gain. Left alone deliberately.

### Verification

- The four paginator tests fail against the whole-block gate and pass after the split.
- The two reproduction cases also fail against `main`, so they still cover the original #44 bug.
- Full suite: 1153 tests across 77 files, all passing. `tsc --noEmit` clean.

src/pages/links.tsx
```diff
@@ -300,15 +300,15 @@ export const LinksPage: FC<Props> = ({
...
-              </div>
+          <div class="pagination">
+            <div class="pagination-summary">
+              {t("links.pageSummary", {
+                from: start + 1,
+                to: Math.min(start + perPage, sorted.length),
+                total: sorted.length,
+              })}
+            </div>
+            {totalPages > 1 && (
```
DennisAlund, line 311: This is the corrected gate. An earlier revision of this PR gated the entire `.pagination` block here on `totalPages > 1`, which fixed #44's dead control but hid two things that are not dead: the result summary and the per-page selector.

The failure it created: 30 active links, user picks 50 from the selector. `totalPages` becomes `ceil(30/50)` = 1, so the whole block vanished, selector included. `buildUrl` hardcodes `per_page: String(perPage)` (`:116`) and the range picker preserves it (`:145`), so every sort link and filter chip carried `per_page=50` forward. `per_page` is URL-only (`src/index.tsx:235`), so the only route back to 25 was leaving the links page entirely.

Splitting the gate keeps `.pagination` tied to the table and scopes `totalPages > 1` to the chevrons and page numbers. #44 asked for the dead control to go, not the summary and the selector.

src/pages/links.tsx
```diff
@@ -300,15 +300,15 @@ export const LinksPage: FC<Props> = ({
...
-                {t("links.pageSummary", {
-                  from: sorted.length === 0 ? 0 : start + 1,
-                  to: Math.min(start + perPage, sorted.length),
-                  total: sorted.length,
-                })}
-              </div>
+          <div class="pagination">
+            <div class="pagination-summary">
+              {t("links.pageSummary", {
+                from: start + 1,
```
DennisAlund, line 306: Collapsed from `sorted.length === 0 ? 0 : start + 1`. The zero branch was already unreachable: `totalPages > 1` implies `ceil(sorted.length / perPage) >= 2`, and `perPage` is clamped to `>= 1` at `src/index.tsx:235`, so `sorted.length > perPage >= 1`.

It was reachable under the old `links.length > 25` escape hatch, which rendered the block with `sorted.length === 0`. That hatch is gone, so the ternary went with it.

src/pages/links.tsx
```diff
@@ -340,25 +340,25 @@ export const LinksPage: FC<Props> = ({
...
-                    {[25, 50, 100].map((n) => (
-                      <option value={perPageUrl(n)} selected={perPage === n}>
-                        {n}
-                      </option>
-                    ))}
-                  </select>
-                </div>
-                <span class="per-page-label">{t("links.perPage")}</span>
+            )}
+            <div class="per-page">
```
DennisAlund, line 344: The selector sits outside the `totalPages` conditional on purpose. It is the only in-page control that writes `per_page`, so gating it on the value it sets makes wider page sizes a one-way door. Worth keeping in mind if this block gets restructured again.

src/__tests__/page/links-page.test.ts
```diff
@@ -224,4 +224,79 @@ describe("Links listing page", () => {
...
+
+  it("hides the page buttons when a wider per_page fits every link on one page", async () => {
+    for (let i = 0; i < 30; i++) {
+      await LinkRepository.create(env.DB, {
+        url: `https://example${i}.com`,
+        slug: `s${i}`,
+      });
+    }
+    const res = await SELF.fetch(req("/_/admin/links?per_page=100"));
+    expect(res.status).toBe(200);
```
DennisAlund, line 236: The status check and the row count are the point here. As first written, both hide-tests asserted only absences (`not.toContain('class="pagination"')`, `not.toMatch(/class="page-btn/)`), which stays green if the page 500s or falls through to the empty state, where no paginator renders by construction. Neighbouring tests at `:27` and `:45` already assert `200`.

Now each case pins a 200, counts rendered rows (30 here, 2 for the filter case), and asserts the summary and selector survive. Non-vacuity is checked both directions: all four fail against the whole-block gate, and the two reproduction cases also fail against `main`.

One trap worth recording: the row class is built from a template string (`col-short-chip no-row-nav...`), so a `class="col-short-chip"` exact match never fires, and a bare `col-short-chip-slug` match counts one extra because the stylesheet is inlined into the page. The assertion anchors on `class="col-short-chip-slug"`.

src/__tests__/page/links-page.test.ts
```diff
@@ -224,4 +224,79 @@ describe("Links listing page", () => {
...
+    const res = await SELF.fetch(req("/_/admin/links"));
+    expect(res.status).toBe(200);
+    const html = await res.text();
+    expect(html).toContain('class="pagination"');
+    expect(html).toContain('class="pagination-pages"');
+    expect(html).toMatch(/class="page-btn[^"]*"[^>]*>2</);
+    expect(html).toContain("per-page-select");
+  });
+
+  it("keeps the per-page selector reachable after widening per_page", async () => {
```
DennisAlund, line 287: Added because the selector's disappearance had no guard rail in either direction. This pins the route back to 25 after widening `per_page`, which is the exact path that broke under the whole-block gate.

src/pages/links.tsx
DennisAlund: Verified against the current head. The gate splits at `:311`: `.pagination` renders whenever the table renders, `totalPages > 1` scopes only `.pagination-pages`.

Reconstructed the whole-block gate locally and re-ran the file: `keeps the per-page selector reachable after widening per_page` fails under it, so the trap this comment describes is now pinned by a test rather than by prose alone.

src/pages/links.tsx
DennisAlund: The collapse is safe, but the invariant cited here no longer carries it. After the split, `totalPages > 1` does not gate the summary, so it cannot rule out `sorted.length === 0`.

The reason it holds now is narrower: this whole block sits inside the `filtered.length > 0` branch (`src/pages/links.tsx:218`), and `sorted` is a copy of `filtered` (`:100`). A zero-length `sorted` renders the empty state instead, so the summary never sees it.

Recording the real invariant matters for the next edit: move `.pagination` out of that branch and `from` reads `1` over a `0` total.

src/pages/links.tsx
DennisAlund: Confirmed at `:344`: the selector sits outside the `totalPages > 1` conditional, and `keeps the per-page selector reachable after widening per_page` (`src/__tests__/page/links-page.test.ts:287`) fails if it moves back inside.

src/__tests__/page/links-page.test.ts
DennisAlund: Re-ran both mutations rather than trusting the note. One correction to the count.

- Against `main`: `hides the page buttons when a wider per_page fits every link on one page` and `hides the page buttons when a filter narrows the result set to one page` fail, 17 pass. The #44 coverage holds.
- Against a reconstructed whole-block gate: three fail, not four. Those two plus `keeps the per-page selector reachable after widening per_page`. The fourth, `still renders the page buttons when the result set spans more than one page`, exercises `totalPages > 1`, where both gate shapes render the block, so it passes either way. It guards the multi-page case, not the split.

The PR description carries the same off-by-one and is corrected there.

src/__tests__/page/links-page.test.ts
DennisAlund: Confirmed non-vacuous in both directions: the case passes on head and fails against a reconstructed whole-block gate.

Dropped a stray blank line that followed it before the closing `describe` (b936d48).

PR #49, opened 2026-08-26
Give the links paginator an accessible name, state, and landmark

### LINKED ISSUE #43: Links paginator controls are inaccessible to keyboard and screen readers
Surfaced while reviewing #41. All three problems predate that PR.

## 1. Prev/next chevrons have no accessible name

`src/pages/links.tsx:272` and `src/pages/links.tsx:288` render:

```html
<a class="page-btn" href="..."><span class="icon icon-sm">chevron_left</span></a>
```

The icon font draws a glyph, but the accessibility tree still holds the literal string, so VoiceOver and NVDA announce "link, chevron_left". The icon span also lacks `aria-hidden`.

Translations already exist and go unused: `linkDetail.prevPage` and `linkDetail.nextPage` at `src/i18n/en.ts:114-115`, plus the `id` and `sv` locales.

## 2. Disabled chevrons stay keyboard-focusable

`.page-btn.disabled` (`src/styles.ts:485`) sets `pointer-events: none`, which blocks the mouse and not the keyboard. On page 1 the prev control renders as `<a class="page-btn disabled" href="#">`. A keyboard user tabs onto a control drawn at 30% opacity, presses Enter, and gets a new history entry plus a jump to the top of the document with no page change. The element carries no `aria-disabled`, so it announces as an ordinary enabled link.

Fix: drop `href` on the disabled state, or add `tabindex="-1"` and `aria-disabled="true"`.

## 3. No landmark or group name around the page links

`.pagination-pages` is a bare `<div>` holding the chevrons and page anchors. A screen reader user hits a run of links announced as "1", "2", "3", "188" with nothing to jump to or past, and `aria-current="page"` (added in #41) has no enclosing navigation context to be current within.

The same file already models the pattern at `src/pages/links.tsx:143`:

```jsx
<div class="filter-chips" role="group" aria-label={t("links.filter")}>
```

Wrapping the paginator in `<nav aria-label={t(...)}>` reuses that convention. New label copy goes through `t()` and the locale files.

DennisAlund:
Closes #43.

Three faults in the links paginator, all predating #41.

## 1. The chevrons had no accessible name

The anchors held the literal string `chevron_left` / `chevron_right`. The icon font draws a glyph, but the accessibility tree keeps the text, so VoiceOver and NVDA announced "link, chevron_left".

The glyph span now carries `aria-hidden="true"` and the control carries a translated `aria-label`, reusing the existing `linkDetail.prevPage` and `linkDetail.nextPage` strings rather than adding listing-page duplicates of the same two words.

## 2. Disabled chevrons stayed keyboard-focusable

`.page-btn.disabled` sets `pointer-events: none`, which stops the mouse and not the Tab key, and the element still rendered `href="#"`. On page 1 a keyboard user tabbed onto a control drawn at 30% opacity, pressed Enter, and got a history entry plus a jump to the top of the document with no page change. The element carried no `aria-disabled`, so it announced as an ordinary enabled link.

The disabled state now renders as a `<span>` with no `href`: no href means no tab stop and no dead navigation, and `aria-disabled="true"` with `role="link"` announces why the control is dimmed.

## 3. No landmark around the page links

`.pagination-pages` was a bare `<div>`. A screen reader user hit a run of links announced as "1", "2", "3", "188" with nothing to jump to or past, and the `aria-current="page"` added in #41 had no navigation context to be current within.

The run now sits in `<nav class="pagination-pages" aria-label={t("links.pagination")}>`, matching the breadcrumb convention in `src/components/topbar.tsx`. The new `links.pagination` key ships in all three locales.

## Implementation note

Both chevrons share a `PageStep` component, so the enabled and disabled shapes cannot drift apart again. The mobile scroll rules from #47 (`.pagination-pages > :first-child` / `:last-child` auto margins) apply to the `<nav>` unchanged.

## Testing

Four tests added to `src/__tests__/page/links-page.test.ts`, one per fault plus the last-page case. Each fails on `main` and passes here. Full suite: 77 files, 1153 tests, all passing. `tsc --noEmit` is clean. No API or spec surface changes, so no SDK hash bump.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NReFFDuZApXowxPCzR1owG)_

### DIFF
--- a/src/__tests__/page/links-page.test.ts
+++ b/src/__tests__/page/links-page.test.ts
@@ -295,6 +295,61 @@ describe("Links listing page", () => {
     expect(pageTwo).not.toContain("search=");
   });
 
+  it("gives every row an anchor to its detail page so the keyboard can reach it", async () => {
+    const link = await LinkRepository.create(env.DB, {
+      url: "https://example.com/some/deep/path",
+      slug: "abc",
+      label: "Example link",
+    });
+
+    const html = await (await SELF.fetch(req("/_/admin/links"))).text();
+
+    // A real anchor is a tab stop and answers Enter. The row onclick is a
+    // mouse convenience and reaches neither.
+    const row = html.match(
+      new RegExp(`<a\\b([^>]*href="/_/admin/links/${link.id}"[^>]*)>([\\s\\S]*?)</a>`),
+    );
+    expect(row).not.toBeNull();
+    expect(row![2].replace(/<[^>]*>/g, "").trim()).toBe("Example link");
+    // The row handler bails on this target, so the anchor navigates once
+    // instead of racing `location.href` to the same URL.
+    expect(row![1]).toContain("no-row-nav");
+  });
+
+  it("keeps the disabled badge out of the row anchor's accessible name", async () => {
+    const link = await LinkRepository.create(env.DB, {
+      url: "https://example.com",
+      slug: "gone",
+      label: "Expired link",
+      expiresAt: Math.floor(Date.now() / 1000) - 60,
+    });
+
+    const html = await (
+      await SELF.fetch(req("/_/admin/links?filter=all"))
+    ).text();
+
+    const row = html.match(
+      new RegExp(`<a\\b[^>]*href="/_/admin/links/${link.id}"[^>]*>([\\s\\S]*?)</a>`),
+    );
+    expect(row).not.toBeNull();
+    // "Expired link", not "Expired link Disabled".
+    expect(row![1].replace(/<[^>]*>/g, "").trim()).toBe("Expired link");
+  });
+
+  it("leaves the row onclick in place as the mouse shortcut", async () => {
+    const link = await LinkRepository.create(env.DB, {
+      url: "https://example.com",
+      slug: "abc",
+    });
+
+    const html = await (await SELF.fetch(req("/_/admin/links"))).text();
+
+    // JSX entity-encodes the quotes inside the attribute value.
+    expect(html).toContain(
+      `location.href=&#39;/_/admin/links/${link.id}&#39;`,
+    );
+  });
+
   it("clamps a negative page param instead of producing an empty, out-of-range slice", async () => {
     for (let i = 0; i < 30; i++) {
       await LinkRepository.create(env.DB, {
--- a/src/__tests__/page/links-page.test.ts
+++ b/src/__tests__/page/links-page.test.ts
@@ -213,6 +213,72 @@ describe("Links listing page", () => {
     expect(html).not.toMatch(/class="page-btn[^"]*"[^>]*>6</);
   });
 
+  it("gives the prev/next chevrons accessible names and hides the glyph text", async () => {
+    for (let i = 0; i < 8; i++) {
+      await LinkRepository.create(env.DB, {
+        url: `https://example${i}.com`,
+        slug: `s${i}`,
+      });
+    }
+    const res = await SELF.fetch(req("/_/admin/links?per_page=1&page=4"));
+    const html = await res.text();
+    expect(html).toMatch(/<a[^>]*class="page-btn"[^>]*aria-label="Previous page"/);
+    expect(html).toMatch(/<a[^>]*class="page-btn"[^>]*aria-label="Next page"/);
+    // The icon-font glyph is decorative and must not reach the accessibility tree.
+    expect(html).toMatch(/<span class="icon icon-sm" aria-hidden="true">chevron_left<\/span>/);
+    expect(html).toMatch(/<span class="icon icon-sm" aria-hidden="true">chevron_right<\/span>/);
+  });
+
+  it("keeps the disabled prev control out of the tab order on the first page", async () => {
+    for (let i = 0; i < 8; i++) {
+      await LinkRepository.create(env.DB, {
+        url: `https://example${i}.com`,
+        slug: `s${i}`,
+      });
+    }
+    const res = await SELF.fetch(req("/_/admin/links?per_page=1&page=1"));
+    const html = await res.text();
+    const prev = html.match(/<[a-z]+[^>]*class="page-btn disabled"[^>]*aria-label="Previous page"[^>]*>/);
+    expect(prev).not.toBeNull();
+    const tag = prev![0];
+    // No href means no keyboard focus and no dead navigation to "#".
+    expect(tag).not.toContain("href");
+    expect(tag).toContain('aria-disabled="true"');
+    expect(tag).toContain('role="link"');
+    // The next control is still live on page 1.
+    const next = html.match(/<a[^>]*aria-label="Next page"[^>]*>/);
+    expect(next).not.toBeNull();
+    expect(next![0]).toContain("href=");
+  });
+
+  it("keeps the disabled next control out of the tab order on the last page", async () => {
+    for (let i = 0; i < 8; i++) {
+      await LinkRepository.create(env.DB, {
+        url: `https://example${i}.com`,
+        slug: `s${i}`,
+      });
+    }
+    const res = await SELF.fetch(req("/_/admin/links?per_page=1&page=8"));
+    const html = await res.text();
+    const next = html.match(/<[a-z]+[^>]*class="page-btn disabled"[^>]*aria-label="Next page"[^>]*>/);
+    expect(next).not.toBeNull();
+    expect(next![0]).not.toContain("href");
+    expect(next![0]).toContain('aria-disabled="true"');
+  });
+
+  it("wraps the page links in a named navigation landmark", async () => {
+    for (let i = 0; i < 8; i++) {
+      await LinkRepository.create(env.DB, {
+        url: `https://example${i}.com`,
+        slug: `s${i}`,
+      });
+    }
+    const res = await SELF.fetch(req("/_/admin/links?per_page=1&page=1"));
+    const html = await res.text();
+    expect(html).toMatch(/<nav class="pagination-pages" aria-label="Pagination">/);
+    expect(html).not.toMatch(/<div class="pagination-pages"/);
+  });
+
   it("clamps a negative per_page param instead of an inverted slice range", async () => {
     for (let i = 0; i < 30; i++) {
       await LinkRepository.create(env.DB, {
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -71,7 +71,11 @@ const en = {
   "links.show": "Show",
   "links.searchResults": "{count} matching links",
   "links.clearSearch": "Clear",
-  "links.pagination": "Pagination",
+
+  // Paginator (shared by the links listing and the link-detail stat pager)
+  "pagination.landmark": "Pagination",
+  "pagination.prev": "Previous page",
+  "pagination.next": "Next page",
 
   // Link Detail
   "linkDetail.title": "Link Details",
@@ -112,8 +116,6 @@ const en = {
   "linkDetail.linkModes": "Access Method",
   "linkDetail.noData": "No data yet",
   "linkDetail.statPager": "{from}-{to} of {total}",
-  "linkDetail.prevPage": "Previous page",
-  "linkDetail.nextPage": "Next page",
   "linkDetail.clickToCopy": "Click to copy",
   "linkDetail.slugs": "Slugs",
   "linkDetail.clicks": "clicks",
--- a/src/i18n/en.ts
+++ b/src/i18n/en.ts
@@ -71,6 +71,7 @@ const en = {
   "links.show": "Show",
   "links.searchResults": "{count} matching links",
   "links.clearSearch": "Clear",
+  "links.pagination": "Pagination",
 
   // Link Detail
   "linkDetail.title": "Link Details",
--- a/src/i18n/id.ts
+++ b/src/i18n/id.ts
@@ -74,7 +74,11 @@ const id: Translations = {
   "links.show": "Tampilkan",
   "links.searchResults": "{count} tautan cocok",
   "links.clearSearch": "Hapus",
-  "links.pagination": "Paginasi",
+
+  // Paginator (shared by the links listing and the link-detail stat pager)
+  "pagination.landmark": "Paginasi",
+  "pagination.prev": "Halaman sebelumnya",
+  "pagination.next": "Halaman berikutnya",
 
   // Link Detail
   "linkDetail.title": "Detail Tautan",
@@ -115,8 +119,6 @@ const id: Translations = {
   "linkDetail.linkModes": "Metode Akses",
   "linkDetail.noData": "Belum ada data",
   "linkDetail.statPager": "{from}-{to} dari {total}",
-  "linkDetail.prevPage": "Halaman sebelumnya",
-  "linkDetail.nextPage": "Halaman berikutnya",
   "linkDetail.clickToCopy": "Klik untuk menyalin",
   "linkDetail.slugs": "Slug",
   "linkDetail.clicks": "klik",
--- a/src/i18n/id.ts
+++ b/src/i18n/id.ts
@@ -74,6 +74,7 @@ const id: Translations = {
   "links.show": "Tampilkan",
   "links.searchResults": "{count} tautan cocok",
   "links.clearSearch": "Hapus",
+  "links.pagination": "Paginasi",
 
   // Link Detail
   "linkDetail.title": "Detail Tautan",
--- a/src/pages/links.tsx
+++ b/src/pages/links.tsx
@@ -25,19 +25,25 @@ export type PaginationItem = number | "ellipsis";
 
 /**
  * One end of the paginator: a chevron that steps to the previous or next page.
- * The glyph is icon-font text, so it stays out of the accessibility tree and
- * the control carries a translated name instead. At the first or last page the
- * step has nowhere to go, so it drops its href: no href means no tab stop and
- * no dead jump to "#", and aria-disabled announces why it is dimmed.
+ * The glyph is icon-font text, so `aria-hidden` keeps it out of the
+ * accessibility tree and the control carries a translated name instead. At the
+ * first or last page the step has nowhere to go, so it drops its href: no href
+ * means no tab stop and no dead jump to "#", and aria-disabled announces why
+ * it is dimmed.
+ *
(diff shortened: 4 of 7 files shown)


src/pages/links.tsx
```diff
@@ -23,6 +23,39 @@ export type LinksFilter = "active" | "disabled" | "all";
...
+    </span>
+  );
+  return href ? (
+    <a class="page-btn" href={href} aria-label={label}>
+      {glyph}
+    </a>
+  ) : (
+    <span
+      class="page-btn disabled"
+      role="link"
```
DennisAlund, line 56: **The disabled step lands between two accessibility models and gets the benefit of neither.**

Dropping `href` removes the element from the tab order, so a keyboard user never reaches it and never hears the `aria-disabled` the docstring says "announces why it is dimmed". Meanwhile `role="link"` keeps it in the accessibility tree, so a browse-mode reader still meets "link, Previous page" and depends on `aria-disabled` support that NVDA and VoiceOver apply unevenly to `role=link` (solid on `button`, patchy on `link`).

A dimmed chevron pointing nowhere carries nothing a screen reader user can act on, and `aria-current="page"` on the number run already conveys the position. Pick one end:

- treat it as decoration: drop `role`/`aria-disabled`/`aria-label` and put `aria-hidden="true"` on the span, or
- keep it announced: render `<a tabindex="0" aria-disabled="true" aria-label={label}>` with no `href`, so the state reaches the users the attribute exists for.

As written it announces a link that cannot be focused or activated.

src/pages/links.tsx
```diff
@@ -327,13 +360,12 @@ export const LinksPage: FC<Props> = ({
...
             </div>
             {pageCount > 1 && (
-              <div class="pagination-pages">
-                <a
-                  class={`page-btn${currentPage <= 1 ? " disabled" : ""}`}
-                  href={currentPage > 1 ? pageUrl(currentPage - 1) : "#"}
-                >
-                  <span class="icon icon-sm">chevron_left</span>
-                </a>
+              <nav class="pagination-pages" aria-label={t("links.pagination")}>
```
DennisAlund: **The landmark stops short of the control it names.**

`pagination-summary` (line 355, `1-25 of 30`) and the page-size selector (line 392) sit outside the `<nav>`. A reader who jumps to the "Pagination" landmark hears `1 2 3 ... 8` and two chevrons, with no total, no current range, and no page-size control without leaving the landmark again. The count context that makes the number run meaningful is exactly what got left outside.

Moving the `<nav>` up to wrap `div.pagination` (line 354) puts the whole pattern under one accessible name and costs nothing: `.pagination` is the flex container, and the mobile rules on `.pagination-pages` keep working on an inner `<div>`.

src/pages/links.tsx
```diff
@@ -23,6 +23,39 @@ export type LinksFilter = "active" | "disabled" | "all";
...
 export type PaginationItem = number | "ellipsis";
 
+/**
+ * One end of the paginator: a chevron that steps to the previous or next page.
+ * The glyph is icon-font text, so it stays out of the accessibility tree and
+ * the control carries a translated name instead. At the first or last page the
+ * step has nowhere to go, so it drops its href: no href means no tab stop and
+ * no dead jump to "#", and aria-disabled announces why it is dimmed.
+ */
+const PageStep: FC<{ icon: string; label: string; href?: string }> = ({
```
DennisAlund: **`icon` and `label` are two free-form strings that must agree, and nothing enforces it.**

`icon="chevron_left"` with `label={t("linkDetail.nextPage")}` type-checks, renders, and is invisible in review because the glyph is `aria-hidden`. The only symptom is a screen reader announcing the wrong direction.

One prop makes the mismatch unrepresentable:

```tsx
const PageStep: FC<{ direction: "prev" | "next"; label: string; href?: string }>
```

with the glyph derived inside. Two call sites, two fewer chances to get it wrong.

src/pages/links.tsx
```diff
@@ -23,6 +23,39 @@ export type LinksFilter = "active" | "disabled" | "all";
...
+    <span
+      class="page-btn disabled"
+      role="link"
+      aria-disabled="true"
+      aria-label={label}
+    >
+      {glyph}
+    </span>
+  );
+};
```
DennisAlund, line 63: **The accessible shape stays private to this page while the pieces it depends on are exported for sharing.**

`pageWindow` and `paginationItems` are exported from this module precisely so the next caller does not repeat their guards ("the next one will be an htmx partial that does not", line 63). `PageStep`, the `<nav>` wrapper, and the `links.pagination` key are not: they stay module-private in a page file.

So the paginator is now four coupled pieces, two shareable and two not. The htmx partial will import the math and re-derive the markup, and the accessible shape this PR just settled will drift on the copy. `src/components/pagination.tsx` exporting one `Pagination` component (taking `currentPage`, `pageCount`, `pageUrl`, `t`) puts the whole contract in one place and matches where `topbar.tsx` already lives.

src/pages/links.tsx
```diff
@@ -23,6 +23,39 @@ export type LinksFilter = "active" | "disabled" | "all";
 
 export type PaginationItem = number | "ellipsis";
 
+/**
+ * One end of the paginator: a chevron that steps to the previous or next page.
+ * The glyph is icon-font text, so it stays out of the accessibility tree and
```
DennisAlund: **The causality reads backwards.**

Being icon-font text is exactly why the glyph *would* reach the accessibility tree; `aria-hidden="true"` is what keeps it out. As written the sentence claims the fault is the fix.

Suggested: "The glyph is icon-font text, so `aria-hidden` keeps it out of the accessibility tree and the control carries a translated name instead."

src/__tests__/page/links-page.test.ts
```diff
@@ -324,6 +324,72 @@ describe("Links listing page", () => {
...
+  it("gives the prev/next chevrons accessible names and hides the glyph text", async () => {
+    for (let i = 0; i < 8; i++) {
+      await LinkRepository.create(env.DB, {
+        url: `https://example${i}.com`,
+        slug: `s${i}`,
+      });
+    }
+    const res = await SELF.fetch(req("/_/admin/links?per_page=1&page=4"));
+    const html = await res.text();
+    expect(html).toMatch(/<a[^>]*class="page-btn"[^>]*aria-label="Previous page"/);
```
DennisAlund, line 391: **These assertions pin JSX attribute order, not behavior.**

`class="page-btn"[^>]*aria-label=` only matches while `class` precedes `aria-label` in `PageStep`. Reordering the props breaks six assertions across three tests with no change to what renders.

This file already solved that: `anchorHrefs` on line 219 was written to be "independent of attribute order". The disabled-step tests below use the better idiom too, matching the tag once and then checking attributes with `toContain`. Worth applying the same treatment here.

src/__tests__/page/links-page.test.ts
```diff
@@ -324,6 +324,72 @@ describe("Links listing page", () => {
     expect(html).not.toMatch(/class="page-btn[^"]*"[^>]*>6</);
   });
 
+  it("gives the prev/next chevrons accessible names and hides the glyph text", async () => {
+    for (let i = 0; i < 8; i++) {
```
DennisAlund, line 383: **The same eight-link loop is copy-pasted into all four new tests, and it already exists on line 312.**

`seedSearchFixture` on line 233 sets this file's convention for shared fixtures. A `seedLinks(8)` helper drops roughly 28 lines from this hunk.

Three of the four new tests then fetch the identical `?per_page=1&page=1` page and assert on different parts of the same HTML. The landmark test (line 380) asserts one regex against a render the disabled-prev test (line 343) already has in hand, so it could fold in and save a seed plus a request.

src/i18n/en.ts
```diff
@@ -71,6 +71,7 @@ const en = {
   "links.show": "Show",
   "links.searchResults": "{count} matching links",
   "links.clearSearch": "Clear",
+  "links.pagination": "Pagination",
```
DennisAlund: **The paginator's three strings now live in two page namespaces, and neither one is the paginator.**

The landmark name lands in `links.*`; the two control names come from `linkDetail.*` (links.tsx lines 366 and 384). Reusing the link-detail strings avoids duplicate copy, which is right, but it also means the link-detail stat pager now owns half the listing paginator's accessible names: retitling `linkDetail.nextPage` for the stat pager silently retitles this control too.

A `pagination.landmark` / `pagination.prev` / `pagination.next` group keeps the three together, decouples them from either page, and moves with the component when it becomes shared.

src/i18n/id.ts
```diff
@@ -74,6 +74,7 @@ const id: Translations = {
   "links.show": "Tampilkan",
   "links.searchResults": "{count} tautan cocok",
   "links.clearSearch": "Hapus",
+  "links.pagination": "Paginasi",
```
DennisAlund: `Paginasi` is a rare loanword in Indonesian interface copy. `Navigasi halaman` is what Indonesian UIs label this landmark, and it also matches the landmark's role rather than transliterating the English term.

This string reaches screen reader users only, so a weak choice here never surfaces in visual QA.

DennisAlund (commented):
## Code review: paginator accessibility

Extra-high-recall pass over `4f7211a` (the merge resolution) diffed against `main`. The three faults the PR names are genuinely fixed, and the four new tests are real regression tests. `src/__tests__/page/links-page.test.ts` runs 39 passing tests on this commit.

Findings below cover one semantic gap in the disabled state the PR introduces, one scope gap in the new landmark, two keyboard-access bugs this pass walks past in the same paginator, plus reuse, altitude, and copy notes. Two findings sit on lines outside the diff and are posted as file-level comments on `src/pages/links.tsx`.

One finding has no line in this PR at all: `src/client.ts` lines 868 and 876 render the stat pager's chevrons as `<span class="icon">chevron_left</span>` with no `aria-hidden`. The buttons carry `aria-label`, so the accessible name is safe there, but the glyph text still renders verbatim when the icon font fails to load, and it is the pattern the next pager will be copied from. The `aria-hidden` habit this PR establishes belongs there too.

src/pages/links.tsx
DennisAlund, line 1: **Line 396: keyboard users can never select 100 rows per page.**

(Posted as a file comment because the line sits just outside this PR's hunks.)

`onchange="location.href=this.value"` navigates the instant the selection changes. On Windows Chrome and Firefox, arrow keys on a closed `<select>` move the selection and fire `change` on every keypress, so a keyboard user on 25 who presses Down navigates straight to `per_page=50` and never reaches 100. Type-ahead does the same on macOS. There is no way to browse the options without committing to one.

That is a WCAG 3.2.2 (On Input) failure sitting four lines below the `<nav>` this PR adds, inside the same `.pagination` block the PR set out to make accessible. Worth folding into this pass rather than filing behind it: navigate on a deliberate commit (a small Apply button), or render the sizes as links the way the filter chips on line 236 already are.

DennisAlund, line 1: **Line 301: no keyboard path from the listing to any link's detail page.**

(Posted as a file comment because the line sits outside this PR's hunks.)

Rows navigate through `onclick` on the `<tr>`, and nothing inside a row is an anchor to `/_/admin/links/{id}`: the label is a `<div>`, the slug is a copy chip, the clicks and date are spans. The `<tr>` carries no `tabindex` and no role, so it is not focusable and Enter does nothing.

A keyboard-only or screen reader user therefore cannot open a link's detail page from this listing at all. That is a WCAG 2.1.1 (Keyboard) failure, and it outranks the three chevron faults this PR fixes: the paginator now announces itself correctly while the destination it pages through stays unreachable.

Wrapping the label in a real anchor fixes it without touching the layout, since the row `onclick` stays as a mouse convenience:

```tsx
<a class="col-link-label-link" href={href}>{link.label || link.url}</a>
```

Reasonable to split into its own PR, but it belongs in the same issue thread as this accessibility work rather than going unrecorded.

src/pages/links.tsx
DennisAlund: **Retracting this: the suggestion breaks the test in this PR.**

I applied it (moved `<nav aria-label>` up to `.pagination`, reverted `.pagination-pages` to a `<div>`) and ran the file: 38 passed, 1 failed. `wraps the page links in a named navigation landmark` asserts both `<nav class="pagination-pages" aria-label="Pagination">` and `not.toMatch(/<div class="pagination-pages"/)`, so either assertion alone red-lights it. Saying this "costs nothing" was wrong.

The observation still holds: the `1-25 of 30` summary and the page-size control sit outside the landmark. But it is a choice between two shapes with a test moving either way, not a free win, so it should not have been posted as a defect. Current shape stands.

src/pages/links.tsx
DennisAlund: **The headline is wrong. Withdrawing it.**

`Alt+Down` opens the listbox, and arrowing inside an *open* listbox does not fire `change` in Chrome or Firefox, so `Enter` on 100 works. On macOS the arrow keys on a closed select open the popup rather than moving the value, so the failure I described never happens there at all. Type-ahead `1` lands on 100, which is the option the user wanted, not a trap. So "can never select 100 rows per page" and "no way to browse the options without committing to one" are both false.

What survives is narrower and pre-existing: navigating on `change` is a context change on input, which is a WCAG 3.2.2 smell. That belongs in its own issue, not folded into this PR as I suggested. Resolving.

src/__tests__/page/links-page.test.ts
DennisAlund: **Correction: two of the four, not three.**

The four new tests fetch `page=4`, `page=1`, `page=8`, `page=1`. Only the disabled-prev and landmark tests share a render, so the comment overstated the duplication it was arguing about.

With the count fixed the residue is a `seedLinks(8)` helper worth roughly 28 lines and one foldable test. Too small to hold up the PR. Resolving.

src/pages/links.tsx
DennisAlund: **Withdrawing the evidence in this comment.**

Two problems with how I argued it:

1. "NVDA and VoiceOver apply `aria-disabled` unevenly to `role=link`, solid on `button`, patchy on `link`" is unsourced. I cannot test AT behavior here, and MDN lists `link` among the roles that support `aria-disabled`. That clause should not have been stated as fact.
2. "A keyboard user never reaches it and never hears the `aria-disabled`" conflates two populations. A sighted keyboard-only user hears nothing either way and reads the 30% opacity. An AT user in browse mode traverses the buffer rather than the tab order, so they do reach the span and do hear the state.

I also failed to say that the `tabindex="0"` option reverses the fix this PR set out to make, which makes it a poor thing to offer as a symmetric alternative.

What is left is a judgment call you already made deliberately, not a defect: the span announces as a link that cannot be activated, and dropping the tab stop is the trade you chose. Resolving.

src/__tests__/page/links-page.test.ts
DennisAlund: **The core point stands; the evidence I cited does not.**

`anchorHrefs` extracts **hrefs**. It cannot assert `aria-label` or `aria-hidden`, so it is not the precedent to apply here as I claimed. Worse, the disabled-step tests I held up as "the better idiom" anchor on `/class="page-btn disabled"[^>]*aria-label="Previous page"/`, which pins attribute order in exactly the way this comment objects to. I praised code sharing the flaw I was reporting.

The underlying fact is verified: six assertions across three tests break if `PageStep` reorders its props, with no change to what renders. But I have no remedy to offer that this file already demonstrates, and test brittleness on a component with two call sites is not worth a rewrite now. Resolving.

src/pages/links.tsx
DennisAlund: **Wrong line reference, and this is speculative generality.**

The htmx-partial comment is at line 83, not line 63.

More to the point: no links-listing partial exists yet. `src/admin/widgets/` holds the dashboard slice only. Proposing `src/components/pagination.tsx` before the second caller exists is the same speculative extraction I would push back on from anyone else, and the right time to make the shape shareable is when the partial lands and can be built against it. Resolving; worth raising again on that PR.

src/i18n/id.ts
DennisAlund: **Reframing this as a question, since I have no standing to assert it.**

I wrote "`Navigasi halaman` is what Indonesian UIs label this landmark" as if it were established. I have no basis for that claim and should not have stated it as fact.

The question is genuine, so leaving this open: for a screen-reader-only landmark name, do you want `Paginasi` or `Navigasi halaman`? Your call.

src/pages/links.tsx
DennisAlund: Fixed in `9953358`. The docstring now reads "The glyph is icon-font text, so `aria-hidden` keeps it out of the accessibility tree and the control carries a translated name instead."

src/pages/links.tsx
DennisAlund: Fixed in `9953358`. `PageStep` now takes `direction: "prev" | "next"` and derives the glyph through a `PAGE_STEP_ICON` lookup, so a chevron cannot point one way while its name says the other. Both call sites pass `direction` and a `label`, and the pairing is no longer theirs to get wrong.

src/i18n/en.ts
DennisAlund: Fixed in `0d23354`. Added `pagination.landmark` / `pagination.prev` / `pagination.next` across all three locales and moved both consumers onto them: the listing paginator here and the link-detail stat pager at `client.ts:868,874`. `links.pagination`, `linkDetail.prevPage` and `linkDetail.nextPage` are gone, so the coupling this comment describes cannot come back. Copy is unchanged in every locale.

src/pages/links.tsx
DennisAlund: Fixed in `55025f8`, in this branch rather than a follow-up.

The label is now a real anchor to `/_/admin/links/{id}`, so Tab reaches it and Enter opens it. Three details worth flagging:

- The disabled badge stays outside the anchor, so the accessible name is `Expired link`, not `Expired link Disabled`.
- The anchor carries `no-row-nav`, the marker the row handler already checks for the copy chip. Without it the row would race `location.href` to the URL the anchor is navigating to anyway.
- It picks up a `:focus-visible` outline in `styles.ts`, since it is now a tab stop that has to show where focus landed.

Three tests cover it: the anchor exists and names the label, the badge stays out of the name, and the row `onclick` survives as the mouse shortcut. Suite is 83 files / 1242 tests green, `tsc --noEmit` clean.

This does widen the PR past the paginator. Say the word and I will lift the commit onto its own branch.

PR #50, opened 2026-08-26
Window the links listing in SQL instead of loading the catalog

### LINKED ISSUE #45: Links page loads and delta-enriches the whole catalog on every request
Surfaced while reviewing #41. That PR fixes the paginator layout at 4,700 links. The request cost behind the same scenario stays untouched, and it is the next thing to break.

## Path

`src/index.tsx:231` calls `listLinks`, which runs `LinkRepository.list` (`src/db/link-repository.ts:31`):

```ts
const links = await db.prepare("SELECT * FROM links ORDER BY created_at DESC").all<Link>();
const slugs = await db.prepare(`SELECT ... FROM slugs s ORDER BY is_custom ASC, created_at ASC`).all<Slug>();

return (links.results ?? []).map((link) => {
  const linkSlugs = (slugs.results ?? []).filter((s) => s.link_id === link.id);
  return assembleLink(link, linkSlugs);
});
```

No `LIMIT`. Both tables load whole, and the `.filter()` inside the `.map()` rescans every slug per link, so assembly is O(links x slugs). `listLinks` then hands the full set to `ClickRepository.attachLinkDeltasBulk` (`src/db/click-repository.ts:763`), which computes deltas for every row. `LinksPage` sorts in JS and slices 25 out at `src/pages/links.tsx:67`.

At 4,700 links every page view ships 4,700 rows and 4,700 delta computations through the Worker to render 25 of them. The symptom shifts from layout overflow to CPU and subrequest limits.

## Direction

`LinkRepository.recent` (`src/db/link-repository.ts:49`) already demonstrates the bounded shape: a `LIMIT`ed link query, then a second query fetching slugs for only those ids, bucketed into a `Map` keyed by `link_id` so assembly stays O(links + slugs). Its own docstring names `list()` as the counterexample.

Pushing the window into SQL means `LIMIT ? OFFSET ?` plus a `COUNT(*)` to derive `totalPages`. Sorting and the active/disabled filter both move into the query, since neither can run in JS once the page holds a slice instead of the catalog.

Worth sequencing against the admin widget-island migration, which will re-shape this route into an htmx partial.

DennisAlund:
Closes [#45](https://github.com/oddbit/shrtnr/issues/45).

## What was happening

`/_/admin/links` read the whole catalog on every request. `LinkRepository.list` ran an unbounded `SELECT * FROM links` plus an unbounded slug query, then rescanned the full slug set once per link, so assembly was O(links x slugs). `listLinks` handed the entire result to `ClickRepository.attachLinkDeltasBulk`, which computed a delta for every row. `LinksPage` then filtered, sorted and sliced 25 rows out in JSX. At 4,700 links that is 4,700 rows and 4,700 delta computations to render 25.

The search path on the same route was worse: `LinkRepository.search` fanned out one `getById` per match, so a broad search spent a subrequest per matching link.

## What changed

`LinkRepository.page()` puts the window in SQL. Three statements whatever the catalog size:

1. `count()` for the total the page was cut from, which is what `totalPages` needs. `page()` calls it rather than building a second `COUNT`, so the toolbar total and the empty-state count cannot diverge.
2. The windowed link rows: `LIMIT ? OFFSET ?`.
3. The slugs of only those rows, bound to the ids the window returned and bucketed into a `Map` keyed by `link_id` so assembly is O(links + slugs).

Sorting and the active/disabled filter moved into the query, since neither can run in JS once the page holds a slice. Popular ordering reuses the same click filters and `sinceTs` as the rendered `click_count`: both count fragments append one shared `clickWindowSql`, so the ranking and the printed numbers cannot drift apart.

`migrations/0007` adds `idx_links_created_at ON links(created_at DESC, id DESC)`. Without it the windowed query still reported `SCAN l | USE TEMP B-TREE FOR ORDER BY`, so SQLite read and sorted the whole table to hand back 25 rows and the route kept paying for the catalog. The index lets the planner walk in order and stop after `limit + offset` rows. A plan test asserts the temp B-tree is gone and fails if the migration is removed.

`LinkRepository.rows()` is the unbounded sibling of `page()`, for the API and MCP callers that return whole sets. `list`, `search`, `findByOwner` and `findByUrl` all delegate to it, and all four derive their predicate from the same `linkFilterSql` that `page()` uses. That gives search, owner and status one SQL definition instead of two divergent ones, and it removes the `Promise.all(ids.map(getById))` fan-out from every caller, not just the admin page. Each costs two statements whatever the match count.

`attachLinkDeltasBulk` now restricts its group-by to the ids it was handed. Clicks outgrow links, so grouping every click in the range to label 25 rows was the largest scan left in the request. The two window bounds moved inline, the way `clickWindowSql` already inlines `sinceTs`, which frees the whole parameter budget for ids and lets a full 100-row page stay inside it. Past the cap the ids will not bind and a caller passing the whole catalog wants the full aggregate anyway, so that path keeps the unrestricted query.

`listLinksPage()` owns the page arithmetic and every bound on `page` and `per_page`; the route only parses them. It returns `LinksPageData` rather than a `ServiceResult`, since it clamps every input and has no failure to report. `perPage` clamps to `[1, LINKS_MAX_PER_PAGE]`, which is now derived from `LINKS_PER_PAGE_OPTIONS` and pinned under `D1_MAX_BOUND_PARAMS` by a test, so growing the selector cannot outrun the slug query's binds. An out-of-range `page` still lands on the last populated window.

`LinksPage` became presentational. It receives the rows it renders plus `total`, `totalPages` and `emptyReason`, and `pageWindow()` floors all three page inputs the way `paginationItems` already floored its own. `emptyReason` carries the one thing the component cannot see for itself, and it now has three cases rather than two: `no-links`, `all-disabled` and `no-matches`. The service picks between them from an unfiltered catalog count, keyed off the served window rather than the total.

## Behavior

Query params, URLs and copy are unchanged, legacy `show_disabled=1` included. Four deliberate differences:

- **Tie ordering on the API and MCP surfaces changes, and the index is what changes it.** With three links sharing a `created_at` second: pre-branch `ORDER BY created_at DESC` planned as `SCAN links | USE TEMP B-TREE FOR ORDER BY` and returned them oldest first; with `idx_links_created_at` the same SQL plans as an index walk and returns them newest first. The explicit `created_at DESC, id DESC` does not introduce that flip, it pins it. Without the tie-break the index would have reordered API and MCP results with nothing in the code saying so, and it would flip back the day the planner stops choosing that index. Nothing depends on the ordering: the spec's only ordering promise is `BreakdownPage.items`, no SDK README or MCP tool description mentions link ordering, and `GET /api/links` takes only `owner` and `range` with no limit/offset, so unstable ties were producing nondeterministic output rather than corrupting pagination. Observable only for links created in the same second.
- `?filter=disabled` on an all-active catalog used to render "All links are disabled. Toggle Show disabled to see them." It now renders "No links match the current filter." The old copy stated the inverse of the truth and named a toggle the filter chips replaced; the dead `links.showDisabled` key is gone from all three locales.
- `per_page` above 100 clamps to 100. Previously it was unbounded.
- An unrecognized `sort` value now renders as `recent` and marks that chip active, instead of marking neither.

## Tests

38 repository cases across `link-page` (windowing, ordering, the status filter, search, counts, one statement-count test pinning three statements at 5 links and at 60, one asserting the window is computed once), `link-query-cost` (two statements per call for `list`, `search`, `findByOwner` and `findByUrl` at 3 matches and at 40) and `link-page-plan` (the index walk). 14 service cases for clamping, page arithmetic and the three empty reasons. 35 page cases asserting the route renders one window, pages the filtered set, and never claims every link is disabled when none is. 19 unit cases for the page window, the per-page bounds and the shared click-count tail. Full suite: 83 files, 1235 tests, all passing. `tsc --noEmit` clean. `./scripts/spec-hash.sh` matches the three recorded SDK hashes, so no spec drift and no SDK work.

## Follow-ups left alone

Sequencing against the admin widget-island migration, as the issue notes, is untouched: the page is now a pure function of a served window, which is the shape an htmx partial wants.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RApYtwk8yHbC9UNAm4ehMX)_

### DIFF
--- a/src/__tests__/repository/link-page.test.ts
+++ b/src/__tests__/repository/link-page.test.ts
@@ -23,7 +23,12 @@ async function click(slug: string, at: number): Promise<void> {
   ).bind(slug, at).run();
 }
 
-/** Counts prepare() calls so a test can assert cost does not scale with catalog size. */
+/**
+ * Counts the statements a call issues so a test can pin the cost of a page
+ * render. Traps exec() alongside prepare(): batch() composes already-prepared
+ * statements and so is counted through prepare, but exec() takes raw SQL and
+ * would otherwise slip past the count entirely.
+ */
 function countingDb(counter: { n: number }): D1Database {
   return new Proxy(env.DB, {
     get(target, prop, receiver) {
@@ -31,6 +36,12 @@ function countingDb(counter: { n: number }): D1Database {
         return (sql: string) => {
           counter.n++;
           return (target as unknown as D1Database).prepare(sql);
+        };
+      }
+      if (prop === "exec") {
+        return (sql: string) => {
+          counter.n++;
+          return (target as unknown as D1Database).exec(sql);
         };
       }
       const value = Reflect.get(target, prop, receiver);
@@ -72,7 +83,7 @@ describe("LinkRepository.page windowing", () => {
     expect((await LinkRepository.page(env.DB, { limit: -3 })).links).toEqual([]);
   });
 
-  it("costs the same number of queries at 5 links as at 60", async () => {
+  it("costs three statements at 5 links and the same three at 60", async () => {
     await seed(5, "a");
     const small = { n: 0 };
     await LinkRepository.page(countingDb(small), { limit: 25 });
@@ -81,7 +92,11 @@ describe("LinkRepository.page windowing", () => {
     const large = { n: 0 };
     const result = await LinkRepository.page(countingDb(large), { limit: 25 });
 
-    expect(large.n).toBe(small.n);
+    // Pin the absolute count, not just the equality: a regression to thirty
+    // statements per render is equally flat across catalog sizes and would
+    // satisfy an equality assertion on its own.
+    expect(small.n).toBe(3);
+    expect(large.n).toBe(3);
     expect(result.links).toHaveLength(25);
     expect(result.total).toBe(60);
   });
--- a/src/__tests__/repository/link-page.test.ts
+++ b/src/__tests__/repository/link-page.test.ts
@@ -24,23 +24,23 @@ async function click(slug: string, at: number): Promise<void> {
 }
 
 /**
- * Counts the statements a call issues so a test can pin the cost of a page
- * render. Traps exec() alongside prepare(): batch() composes already-prepared
- * statements and so is counted through prepare, but exec() takes raw SQL and
- * would otherwise slip past the count entirely.
+ * Records the SQL of every statement a call issues, so a test can pin both the
+ * cost of a page render and the shape of what it sends. Traps exec() alongside
+ * prepare(): batch() composes already-prepared statements and so is recorded
+ * through prepare, but exec() takes raw SQL and would slip past entirely.
  */
-function countingDb(counter: { n: number }): D1Database {
+function spyDb(log: string[]): D1Database {
   return new Proxy(env.DB, {
     get(target, prop, receiver) {
       if (prop === "prepare") {
         return (sql: string) => {
-          counter.n++;
+          log.push(sql);
           return (target as unknown as D1Database).prepare(sql);
         };
       }
       if (prop === "exec") {
         return (sql: string) => {
-          counter.n++;
+          log.push(sql);
           return (target as unknown as D1Database).exec(sql);
         };
       }
@@ -85,20 +85,46 @@ describe("LinkRepository.page windowing", () => {
 
   it("costs three statements at 5 links and the same three at 60", async () => {
     await seed(5, "a");
-    const small = { n: 0 };
-    await LinkRepository.page(countingDb(small), { limit: 25 });
+    const small: string[] = [];
+    await LinkRepository.page(spyDb(small), { limit: 25 });
 
     await seed(55, "b");
-    const large = { n: 0 };
-    const result = await LinkRepository.page(countingDb(large), { limit: 25 });
+    const large: string[] = [];
+    const result = await LinkRepository.page(spyDb(large), { limit: 25 });
 
     // Pin the absolute count, not just the equality: a regression to thirty
     // statements per render is equally flat across catalog sizes and would
     // satisfy an equality assertion on its own.
-    expect(small.n).toBe(3);
-    expect(large.n).toBe(3);
+    expect(small).toHaveLength(3);
+    expect(large).toHaveLength(3);
     expect(result.links).toHaveLength(25);
     expect(result.total).toBe(60);
+  });
+
+  it("computes the filtered, sorted window exactly once per render", async () => {
+    await seed(30);
+    const log: string[] = [];
+    await LinkRepository.page(spyDb(log), { limit: 5, offset: 5, sort: "popular" });
+
+    // Under sort=popular the window's ORDER BY is a correlated COUNT over
+    // clicks joined to slugs, so re-running it to fetch the slugs doubles the
+    // per-row aggregates. Only the link query may carry the window, and only
+    // it may carry the link-level aggregate the ordering rests on. The slug
+    // query has its own per-slug click_count, hence the narrower match.
+    expect(log.filter((sql) => sql.includes("LIMIT ? OFFSET ?"))).toHaveLength(1);
+    expect(log.filter((sql) => sql.includes("JOIN slugs cs"))).toHaveLength(1);
+  });
+
+  it("binds the served ids to fetch slugs, staying inside the D1 parameter cap", async () => {
+    await seed(3);
+    const log: string[] = [];
+    const result = await LinkRepository.page(spyDb(log), { limit: 2 });
+
+    const slugQuery = log.find((sql) => sql.includes("FROM slugs s WHERE s.link_id IN"));
+    expect(slugQuery).toBeDefined();
+    expect(slugQuery).toContain("IN (?,?)");
+    expect(result.links).toHaveLength(2);
+    expect(result.links.every((l) => l.slugs.length === 1)).toBe(true);
   });
 
   it("attaches every slug of the windowed links and no others", async () => {
--- a/src/__tests__/repository/link-page.test.ts
+++ b/src/__tests__/repository/link-page.test.ts
@@ -3,7 +3,7 @@
 
 import { beforeAll, beforeEach, describe, expect, it } from "vitest";
 import { env } from "cloudflare:test";
-import { applyMigrations, resetData } from "../setup";
+import { applyMigrations, resetData, spyDb } from "../setup";
 import { LinkRepository, SlugRepository } from "../../db";
 
 beforeAll(applyMigrations);
@@ -21,33 +21,6 @@ async function click(slug: string, at: number): Promise<void> {
   await env.DB.prepare(
     "INSERT INTO clicks (slug, clicked_at, link_mode, is_bot, is_self_referrer) VALUES (?, ?, 'link', 0, 0)",
   ).bind(slug, at).run();
-}
-
-/**
- * Records the SQL of every statement a call issues, so a test can pin both the
- * cost of a page render and the shape of what it sends. Traps exec() alongside
- * prepare(): batch() composes already-prepared statements and so is recorded
- * through prepare, but exec() takes raw SQL and would slip past entirely.
- */
-function spyDb(log: string[]): D1Database {
-  return new Proxy(env.DB, {
-    get(target, prop, receiver) {
-      if (prop === "prepare") {
-        return (sql: string) => {
-          log.push(sql);
-          return (target as unknown as D1Database).prepare(sql);
-        };
-      }
-      if (prop === "exec") {
-        return (sql: string) => {
-          log.push(sql);
-          return (target as unknown as D1Database).exec(sql);
-        };
-      }
-      const value = Reflect.get(target, prop, receiver);
-      return typeof value === "function" ? value.bind(target) : value;
-    },
-  }) as unknown as D1Database;
 }
 
 describe("LinkRepository.page windowing", () => {
--- a/src/__tests__/repository/link-page.test.ts
+++ b/src/__tests__/repository/link-page.test.ts
@@ -1,0 +1,295 @@
+// Copyright 2026 Oddbit (https://oddbit.id)
+// SPDX-License-Identifier: Apache-2.0
+
+import { beforeAll, beforeEach, describe, expect, it } from "vitest";
+import { env } from "cloudflare:test";
+import { applyMigrations, resetData } from "../setup";
+import { LinkRepository, SlugRepository } from "../../db";
+
+beforeAll(applyMigrations);
+beforeEach(resetData);
+
+const NOW = Math.floor(Date.now() / 1000);
+
+async function seed(count: number, prefix = "s"): Promise<void> {
+  for (let i = 0; i < count; i++) {
+    await LinkRepository.create(env.DB, { url: `https://example${i}.com`, slug: `${prefix}${i}` });
+  }
+}
+
+async function click(slug: string, at: number): Promise<void> {
+  await env.DB.prepare(
+    "INSERT INTO clicks (slug, clicked_at, link_mode, is_bot, is_self_referrer) VALUES (?, ?, 'link', 0, 0)",
+  ).bind(slug, at).run();
+}
+
+/** Counts prepare() calls so a test can assert cost does not scale with catalog size. */
+function countingDb(counter: { n: number }): D1Database {
+  return new Proxy(env.DB, {
+    get(target, prop, receiver) {
+      if (prop === "prepare") {
+        return (sql: string) => {
+          counter.n++;
+          return (target as unknown as D1Database).prepare(sql);
+        };
+      }
+      const value = Reflect.get(target, prop, receiver);
+      return typeof value === "function" ? value.bind(target) : value;
+    },
+  }) as unknown as D1Database;
+}
+
+describe("LinkRepository.page windowing", () => {
+  it("returns at most `limit` rows and the unwindowed total", async () => {
+    await seed(30);
+    const result = await LinkRepository.page(env.DB, { limit: 25 });
+    expect(result.links).toHaveLength(25);
+    expect(result.total).toBe(30);
+    expect(result.offset).toBe(0);
+  });
+
+  it("walks disjoint windows across pages", async () => {
+    await seed(30);
+    const first = await LinkRepository.page(env.DB, { limit: 25, offset: 0 });
+    const second = await LinkRepository.page(env.DB, { limit: 25, offset: 25 });
+    expect(second.links).toHaveLength(5);
+    expect(second.offset).toBe(25);
+    const firstIds = new Set(first.links.map((l) => l.id));
+    expect(second.links.some((l) => firstIds.has(l.id))).toBe(false);
+  });
+
+  it("clamps an offset past the end to the last populated window", async () => {
+    await seed(30);
+    const result = await LinkRepository.page(env.DB, { limit: 25, offset: 500 });
+    expect(result.offset).toBe(25);
+    expect(result.links).toHaveLength(5);
+    expect(result.total).toBe(30);
+  });
+
+  it("returns an empty page for a non-positive limit rather than the whole table", async () => {
+    await seed(5);
+    expect((await LinkRepository.page(env.DB, { limit: 0 })).links).toEqual([]);
+    expect((await LinkRepository.page(env.DB, { limit: -3 })).links).toEqual([]);
+  });
+
+  it("costs the same number of queries at 5 links as at 60", async () => {
+    await seed(5, "a");
+    const small = { n: 0 };
+    await LinkRepository.page(countingDb(small), { limit: 25 });
+
+    await seed(55, "b");
+    const large = { n: 0 };
+    const result = await LinkRepository.page(countingDb(large), { limit: 25 });
+
+    expect(large.n).toBe(small.n);
+    expect(result.links).toHaveLength(25);
+    expect(result.total).toBe(60);
+  });
+
+  it("attaches every slug of the windowed links and no others", async () => {
+    const link = await LinkRepository.create(env.DB, { url: "https://example.com", slug: "abc" });
+    await SlugRepository.addCustom(env.DB, link.id, "alias-one");
+    await SlugRepository.addCustom(env.DB, link.id, "alias-two");
+    const other = await LinkRepository.create(env.DB, { url: "https://other.com", slug: "xyz" });
+
+    const result = await LinkRepository.page(env.DB, { limit: 1 });
+    expect(result.links).toHaveLength(1);
+    expect(result.links[0].id).toBe(other.id);
+    expect(result.links[0].slugs.map((s) => s.slug)).toEqual(["xyz"]);
+
+    const secondPage = await LinkRepository.page(env.DB, { limit: 1, offset: 1 });
+    expect(secondPage.links[0].slugs.map((s) => s.slug).sort()).toEqual(["abc", "alias-one", "alias-two"]);
+  });
+});
+
+describe("LinkRepository.page ordering", () => {
+  it("orders by created_at descending with an id tie-break", async () => {
+    for (let i = 0; i < 4; i++) {
+      await LinkRepository.create(env.DB, { url: `https://example${i}.com`, slug: `s${i}` });
+    }
+    // Same second for every row: the id tie-break decides, so a window is stable.
+    await env.DB.prepare("UPDATE links SET created_at = ?").bind(NOW).run();
+    const first = await LinkRepository.page(env.DB, { limit: 2 });
+    const second = await LinkRepository.page(env.DB, { limit: 2, offset: 2 });
+    const ids = [...first.links, ...second.links].map((l) => l.id);
+    expect(ids).toEqual([...ids].sort((a, b) => b - a));
+  });
+
+  it("sorts popular by total clicks across the whole catalog, not just the first window", async () => {
+    await seed(6);
+    // The oldest link (s0) is the most clicked, so recent order would bury it
+    // on the last page while popular order must surface it first.
+    await click("s0", NOW - 60);
+    await click("s0", NOW - 61);
+    await click("s3", NOW - 62);
+
+    const popular = await LinkRepository.page(env.DB, { limit: 2, sort: "popular" });
+    expect(popular.links[0].slugs[0].slug).toBe("s0");
+    expect(popular.links[0].total_clicks).toBe(2);
+    expect(popular.links[1].slugs[0].slug).toBe("s3");
+  });
+
+  it("scopes popular ordering to the requested click window", async () => {
+    await seed(3);
+    // s0 dominates lifetime, s2 dominates the last 7 days.
+    await click("s0", NOW - 60 * 86400);
+    await click("s0", NOW - 61 * 86400);
+    await click("s2", NOW - 3600);
+
+    const lifetime = await LinkRepository.page(env.DB, { limit: 1, sort: "popular" });
+    expect(lifetime.links[0].slugs[0].slug).toBe("s0");
+
+    const week = await LinkRepository.page(
+      env.DB,
+      { limit: 1, sort: "popular" },
+      { sinceTs: NOW - 7 * 86400 },
+    );
+    expect(week.links[0].slugs[0].slug).toBe("s2");
+    expect(week.links[0].total_clicks).toBe(1);
+  });
+
+  it("excludes bot clicks from popular ordering when filtered", async () => {
+    await seed(2);
+    await env.DB.prepare(
+      "INSERT INTO clicks (slug, clicked_at, link_mode, is_bot, is_self_referrer) VALUES (?, ?, 'link', 1, 0)",
+    ).bind("s0", NOW - 60).run();
+    await env.DB.prepare(
+      "INSERT INTO clicks (slug, clicked_at, link_mode, is_bot, is_self_referrer) VALUES (?, ?, 'link', 1, 0)",
+    ).bind("s0", NOW - 61).run();
+    await click("s1", NOW - 62);
+
+    const filtered = await LinkRepository.page(
+      env.DB,
+      { limit: 1, sort: "popular" },
+      { filters: { excludeBots: true } },
+    );
+    expect(filtered.links[0].slugs[0].slug).toBe("s1");
+    expect(filtered.links[0].total_clicks).toBe(1);
+  });
+});
+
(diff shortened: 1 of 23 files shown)


src/db/link-repository.ts
```diff
@@ -27,15 +27,177 @@ function assembleLink(link: Link, slugs: Slug[]): LinkWithSlugs {
...
+    // created_at is second-granularity, so tie-break on id to keep windows
+    // disjoint when several links share a timestamp.
+    const order = query.sort === "popular"
+      ? `${linkClickCountSql(opts)} DESC, l.created_at DESC, l.id DESC`
+      : "l.created_at DESC, l.id DESC";
+    const windowSql = `SELECT l.id FROM links l${where} ORDER BY ${order} LIMIT ? OFFSET ?`;
+
+    const [links, slugs] = await Promise.all([
+      db
+        .prepare(`SELECT l.* FROM links l${where} ORDER BY ${order} LIMIT ? OFFSET ?`)
```
DennisAlund: **No index backs `links(created_at)`, so the windowed query still scans and sorts the whole table.**

`EXPLAIN QUERY PLAN` on the SQL this method builds returns `SCAN l | USE TEMP B-TREE FOR ORDER BY` here, `SCAN l` for the COUNT on line 146, and `SCAN l | USE TEMP B-TREE FOR ORDER BY` again inside the slug subquery on line 175. At 4,700 links, rendering 25 rows still makes D1 read and sort the entire links table three times per request.

The PR removes the JS-side O(links x slugs) assembly and the 4,700 delta computations, so it is a real improvement. The claim that "the route pays for the rows it renders, not for the catalog" is only half true though: the SQL cost stays O(catalog log catalog) per page view.

A migration adding `CREATE INDEX idx_links_created_at ON links(created_at DESC, id DESC)` would let the recent sort walk the index and stop after `limit + offset` rows.

src/db/link-repository.ts
```diff
@@ -27,15 +27,177 @@ function assembleLink(link: Link, slugs: Slug[]): LinkWithSlugs {
...
+        .bind(...binds, limit, offset)
+        .all<Link>(),
+      // The window repeats as a subquery instead of binding the ids it
+      // returned: D1 caps bound parameters per statement, and a 100-row page
+      // would sit at that ceiling.
+      db
+        .prepare(
+          `SELECT ${slugSelect(opts)} FROM slugs s WHERE s.link_id IN (${windowSql}) ORDER BY is_custom ASC, created_at ASC`,
```
DennisAlund: **The window subquery runs a second time here, so the filtered, sorted window is computed twice per request.**

`windowSql` is embedded verbatim inside this `IN (...)`, so SQLite executes the full WHERE + ORDER BY + LIMIT/OFFSET twice on every page render. Under `sort=popular` the ORDER BY is a correlated `COUNT(*)` over clicks joined to slugs, which EXPLAIN shows as one aggregate subquery per catalog row: roughly 9,400 correlated aggregates at 4,700 links instead of 4,700.

The stated reason for the duplication, D1's 100-parameter cap, does not hold at the boundary the comment cites. The limit is 100 bound parameters per statement and a 100-row page needs exactly 100, which is at the cap, not over it. Awaiting the id query once and binding the ids removes the second execution, as does setting `LINKS_MAX_PER_PAGE` below 100.

DennisAlund (commented):
## Review: SQL-windowed links listing

The core mechanism is sound. I checked the riskiest assumption in the design, that SQLite honors `ORDER BY` inside `IN (SELECT ... LIMIT ? OFFSET ?)` for the popular sort at a non-zero offset: it does, and the slug attachment lines up correctly. The new suites pass on the branch (56 tests across the three files) and `tsc --noEmit` is clean.

The headline concern is that the PR's central claim is half-delivered. The JS-side fan-out is gone, but the route still pays catalog-sized costs in SQL, and the fan-out this PR set out to remove still lives on the MCP path.

**Two findings sit on lines this diff does not touch, so they could not be posted inline:**

**`src/db/link-repository.ts:404` - the `getById` fan-out survives for every other search caller.**
`LinkRepository.search` is unchanged apart from the extracted `likePattern`. It still does `Promise.all(ids.map(({ id }) => LinkRepository.getById(db, id, opts)))`, two D1 statements per match. `searchLinks` (src/services/link-management.ts:533) still calls it and is still wired to MCP `search_links` (src/mcp/server.ts:422). A broad MCP search like "http" over 4,700 links issues roughly 9,400 subrequests and trips D1's per-invocation subrequest limit, which is the exact failure this PR set out to fix. The new `linkFilterSql` predicate is a strictly better search implementation; routing `search()` through `page()` or a shared unwindowed variant fixes the fan-out at the mechanism instead of only on the admin page.

**`src/db/link-repository.ts:374` - two divergent SQL definitions of what "search" means.**
`linkFilterSql` matches slugs via `EXISTS (SELECT 1 FROM slugs ms ...)` and orders `created_at DESC, id DESC`; `search()` matches via `LEFT JOIN slugs s` with `SELECT DISTINCT` and orders `created_at DESC` with no id tiebreak. Both are live: the admin page uses the first, the API and MCP the second. Adding a searchable column or changing case handling silently applies to one surface only, and the two already disagree on tie ordering, so the same query returns links in different orders depending on the entry point.

src/services/link-management.ts
```diff
@@ -48,6 +48,85 @@ export async function listLinks(env: Env, opts?: ListLinksOptions): Promise<Serv
...
+    // One extra count, only on the empty path: it separates "no links yet"
+    // from "the status filter hid them all".
+    const unfiltered = status === "all"
+      ? 0
+      : await LinkRepository.count(env.DB, { search: opts?.search, searchOwner: opts?.includeOwner });
+    emptyReason = unfiltered > 0 ? "all-filtered" : "no-links";
+  }
+
+  const links = opts?.withDeltaRange
+    ? await ClickRepository.attachLinkDeltasBulk(env.DB, result.links, opts.withDeltaRange, undefined, opts.filters)
```
DennisAlund, line 140: **`attachLinkDeltasBulk` still aggregates the entire clicks table for the 25 rows actually served.**

The call passes only `result.links` (at most 100 rows), but `attachLinkDeltasBulk` (src/db/click-repository.ts:763) issues two `SELECT s.link_id, COUNT(*) FROM clicks c JOIN slugs s ON c.slug = s.slug WHERE c.clicked_at >= ? GROUP BY s.link_id` queries with no `link_id` restriction.

Clicks grow faster than links. At 4,700 links and 500k clicks, every links-page render groups all 500k rows twice to compute 25 delta pills, which leaves the largest scan in the route untouched by this PR. Restricting the GROUP BY to the served ids (`AND s.link_id IN (...)`, comfortably under the parameter cap at page sizes of 25 or 50) makes the delta cost scale with the window like the rest of the change.

src/pages/links.tsx
```diff
@@ -210,16 +207,16 @@ export const LinksPage: FC<Props> = ({
...
         </div>
       </div>
 
-      {filtered.length === 0 ? (
+      {links.length === 0 ? (
         <div class="empty-state">
           <span class="icon">link_off</span>
           <p>
-            {links.length > 0
+            {emptyReason === "all-filtered"
```
DennisAlund: **The two-valued `emptyReason` cannot express "the disabled filter matched nothing", and that state renders inverted copy.**

Seed 3 active links, request `/_/admin/links?filter=disabled`, and the empty state renders "All links are disabled. Toggle \"Show disabled\" to see them." `listLinksPage` counts the catalog ignoring status, gets 3 > 0, and emits `all-filtered`, which maps to `links.allDisabled` here.

Two problems compound: the sentence states the inverse of the truth (nothing is disabled), and it names a "Show disabled" toggle that the filter-chip toolbar replaced. The behavior predates this PR, but the PR rewrites exactly this decision into a named enum, which is the moment to add the third case (`no-matches`) instead of encoding the old two-way guess.

src/services/link-management.ts
```diff
@@ -48,6 +48,85 @@ export async function listLinks(env: Env, opts?: ListLinksOptions): Promise<Serv
...
+    },
+    { filters: opts?.filters, sinceTs },
+  );
+
+  const totalPages = Math.max(1, Math.ceil(result.total / perPage));
+  // The repository clamps an out-of-range offset, so derive the served page
+  // from what came back rather than from what was asked for.
+  const page = Math.floor(result.offset / perPage) + 1;
+
+  let emptyReason: LinksEmptyReason | undefined;
```
DennisAlund, line 125: **`emptyReason` is computed only when `total === 0`, but the page renders the empty state whenever `links.length === 0`.**

`page()` runs the COUNT and the window as separate statements. If rows are deleted between them, or a request lands on a clamped offset whose rows vanish, `total` comes back positive while `links` is empty. The service then leaves `emptyReason` undefined and `LinksPage` (src/pages/links.tsx:212) falls through its `all-filtered` check to `t("links.empty")`.

The rendered page then reads "30 links" in the toolbar count directly above "No links yet". Deriving `emptyReason` from the served window rather than from `total` closes the gap.

src/pages/links.tsx
```diff
@@ -88,25 +102,8 @@ export const LinksPage: FC<Props> = ({
...
-
-  const sorted = [...filtered].sort((a, b) =>
-    sort === "popular"
-      ? b.total_clicks - a.total_clicks
-      : b.created_at - a.created_at,
-  );
-
-  const totalPages = Math.max(1, Math.ceil(sorted.length / perPage));
-  const currentPage = Math.min(page, totalPages);
+  const currentPage = Math.min(Math.max(1, page), totalPages);
```
DennisAlund: **`currentPage` clamps against `totalPages` without clamping `totalPages` itself.**

`paginationItems` opens with `totalPages = Math.max(1, totalPages)` and the comment "Clamp here rather than at the call site: no caller should have to repeat these guards, and the next one will be an htmx partial that does not." This line then trusts its own `totalPages`.

A caller passing `totalPages: 0` (an htmx partial computing `Math.ceil(0 / perPage)` without the `Math.max(1, ...)` the service applies) yields `currentPage = 0`, then `start = -perPage`, and the summary renders "-24 - 0 of 0" while `pageUrl` emits `page=0`. The PR makes this component presentational and reusable, so its inputs deserve the same clamp its helper already documents.

src/db/link-repository.ts
```diff
@@ -27,15 +27,177 @@ function assembleLink(link: Link, slugs: Slug[]): LinkWithSlugs {
...
+    // must short-circuit rather than reach SQL and return the whole table.
+    const limit = Math.floor(query.limit);
+    if (limit <= 0) return { links: [], total: 0, offset: 0 };
+    // A search that trims to nothing matches everything under LIKE, which is
+    // the opposite of what an empty search box means.
+    if (query.search !== undefined && !query.search.trim()) return { links: [], total: 0, offset: 0 };
+
+    const { where, binds } = linkFilterSql(query);
+    const totalRow = await db
+      .prepare(`SELECT COUNT(*) AS n FROM links l${where}`)
```
DennisAlund: **This COUNT duplicates the statement `count()` already owns.**

This line builds ``SELECT COUNT(*) AS n FROM links l${where}`` and line 199 in `count()` builds the identical string from the identical `linkFilterSql` output, including the same blank-search short-circuit above each.

A change to how totals are computed (adding a soft-delete predicate, switching to `COUNT(l.id)`, counting through a different alias) has to land in two places, and a miss makes `total` and the empty-state count disagree silently. `page()` can call `LinkRepository.count(db, query)` and keep only the binds it needs for the window.

src/constants.ts
```diff
@@ -19,6 +19,13 @@ export const BREAKDOWN_PAGE_SIZE = 10;
 // Upper bound a single breakdown request may fetch.
 export const MAX_BREAKDOWN_LIMIT = 100;
 
+// Links listing page size. The options feed the per-page selector, and the
+// maximum caps what a hand-edited per_page query can pull into one page so the
+// route's cost stays bounded by the window, not by the catalog.
+export const LINKS_PER_PAGE_OPTIONS = [25, 50, 100] as const;
+export const LINKS_DEFAULT_PER_PAGE: (typeof LINKS_PER_PAGE_OPTIONS)[number] = 25;
+export const LINKS_MAX_PER_PAGE = 100;
```
DennisAlund: **`LINKS_MAX_PER_PAGE` hardcodes a value the options array already carries.**

Line 25 declares `[25, 50, 100]` and this line independently declares `100`. Adding 250 to the options makes the selector offer it while `listLinksPage` clamps to 100: the page serves 100 rows, `perPage === n` is false for every option, no `<option selected>` is emitted, and the browser shows the first option (25) next to a 100-row table.

`export const LINKS_MAX_PER_PAGE = Math.max(...LINKS_PER_PAGE_OPTIONS);` makes that drift impossible.

src/index.tsx
```diff
@@ -226,27 +225,42 @@ app.get("/_/admin/links", async (c) => {
...
+    perPage: requestedPerPage,
+    sort,
+    status: filter,
+    search: searchQuery || undefined,
+    includeOwner: true,
+    withDeltaRange: range,
+    filters,
+    range,
+  });
+  const data = pageResult.ok
```
DennisAlund: **This fallback is unreachable and hand-copies the `LinksPageData` shape.**

`listLinksPage` returns `ok(...)` on every path (src/services/link-management.ts:126) and has no `fail` branch, so this arm never executes. It still spells out `{ links: [], total: 0, page: 1, perPage: LINKS_DEFAULT_PER_PAGE, totalPages: 1, emptyReason: "no-links" }`, one more copy of the page-data shape that a future field addition has to remember.

Either drop the `ServiceResult` wrapper and return `LinksPageData` directly, or drop the fallback.

src/services/index.ts
```diff
@@ -19,6 +20,12 @@ export {
   findSlugForRedirect,
   recordClick,
 } from "./link-management";
+export type {
```
DennisAlund: **These four type re-exports have no consumer.**

`ListLinksOptions`, `ListLinksPageOptions`, `LinksPageData` and `LinksEmptyReason` are exported here, while the one consumer reaches past the barrel: `src/pages/links.tsx:11` does `import type { LinksEmptyReason } from "../services/link-management"`. A grep across `src/` finds no other reference to any of the four.

Either point the page at the barrel or drop the entries; as it stands the page stays coupled to the module path and the new exports are dead surface.

src/db/filters.ts
```diff
@@ -48,3 +48,19 @@ export function slugClickCountSql(opts?: SlugClickCountOptions): string {
...
 }
+
+/**
+ * Scalar subquery totalling a link's clicks across all its slugs, under the
+ * same bot/self-referrer filters and `sinceTs` window as
+ * `slugClickCountSql`. Use it to order a link query by popularity in SQL so
+ * the ranking matches the `total_clicks` the caller will render. Pass the
+ * alias the outer query gives the links table.
+ */
+export function linkClickCountSql(opts?: SlugClickCountOptions, linkAlias = "l"): string {
```
DennisAlund, line 69: **`linkClickCountSql` copy-pastes the tail of `slugClickCountSql`.**

Both functions build the same body: open a `SELECT COUNT(*) FROM clicks c`, append `clickFilterSql(opts?.filters, "c")`, then conditionally append `` ` AND c.clicked_at >= ${Math.floor(opts.sinceTs)}` ``. Only the FROM and the correlation differ.

The doc comment states the two must stay under "the same bot/self-referrer filters and `sinceTs` window" for popular ordering to match the rendered `total_clicks`, which is precisely the invariant a copy-paste pair breaks. Factoring the shared tail into one helper both call makes that invariant structural instead of documented.

src/__tests__/repository/link-page.test.ts
```diff
@@ -0,0 +1,295 @@
...
+  it("costs the same number of queries at 5 links as at 60", async () => {
+    await seed(5, "a");
+    const small = { n: 0 };
+    await LinkRepository.page(countingDb(small), { limit: 25 });
+
+    await seed(55, "b");
+    const large = { n: 0 };
+    const result = await LinkRepository.page(countingDb(large), { limit: 25 });
+
+    expect(large.n).toBe(small.n);
```
DennisAlund: **This assertion pins the wrong property.**

`expect(large.n).toBe(small.n)` passes if `page()` regresses from 3 queries to 30, as long as it is 30 at both catalog sizes, so it does not test what the method's docstring claims ("Three queries whatever the catalog size").

Separately, `countingDb` on line 27 traps only `prepare`, so a refactor to `db.batch([...])` would record `n === 0` at both sizes and still pass. Assert `expect(small.n).toBe(3)` alongside the equality, and count `batch` calls in the trap.

src/index.tsx
```diff
@@ -226,27 +225,42 @@ app.get("/_/admin/links", async (c) => {
...
   const range = (validRanges.has(rangeParam as TimelineRange) ? rangeParam : defaultRange) as TimelineRange;
-  const linksResult = searchQuery
-    ? await searchLinks(c.env, searchQuery, { includeOwner: true, withDeltaRange: range, filters, range })
-    : await listLinks(c.env, { withDeltaRange: range, filters, range });
-  const links = linksResult.ok ? linksResult.data : [];
-  const sort = c.req.query("sort") || "recent";
-  const page = Math.max(1, parseInt(c.req.query("page") || "1", 10) || 1);
-  const perPage = Math.max(1, parseInt(c.req.query("per_page") || "25", 10) || 25);
+  const sort = c.req.query("sort") === "popular" ? "popular" : "recent";
+  const requestedPage = Math.max(1, parseInt(c.req.query("page") || "1", 10) || 1);
```
DennisAlund: **Page and per-page clamping is split across the route and the service.**

This line does `Math.max(1, parseInt(...) || 1)` for the page, then `listLinksPage` (src/services/link-management.ts:91) redoes `Math.max(1, Math.floor(opts?.page ?? 1) || 1)`. Line 230 leaves `per_page` unclamped and lets the service handle it.

Two layers own overlapping halves of one rule, so a future change to page-number handling lands in the route or the service depending on which half it touches. The comment on line 238 ("listLinksPage clamps page and per_page") is already only half accurate about the lines above it. Parsing in the route and clamping only in the service settles ownership.

src/db/link-repository.ts
DennisAlund: Fixed in 66c3d10. migrations/0007 adds `CREATE INDEX idx_links_created_at ON links(created_at DESC, id DESC)`.

A plan test asserts the temp B-tree is gone: `EXPLAIN QUERY PLAN` on the window now reports the index walk instead of `SCAN l | USE TEMP B-TREE FOR ORDER BY`. I checked it has teeth by removing the migration file, and both assertions in `link-page-plan.test.ts` fail without it.

src/db/link-repository.ts
DennisAlund: Fixed in b1cee30. The slug query binds the ids the links window returned, the way `recent()` already did, so the window computes once.

You are right that the 100-parameter argument does not hold at the boundary I cited. Since `LINKS_MAX_PER_PAGE` is now derived from `LINKS_PER_PAGE_OPTIONS` (f89f208), a future option above the cap would have broken the slug query silently, so `D1_MAX_BOUND_PARAMS` is now a named constant and a test pins `LINKS_MAX_PER_PAGE` under it.

A test asserts one statement carries `LIMIT ? OFFSET ?` and one carries the link-level aggregate. The slug query legitimately contains `FROM clicks` for its own per-slug `click_count`, so the assertion matches `JOIN slugs cs` instead.

src/services/link-management.ts
DennisAlund: Fixed in 8cc43f3. The group-by restricts to the ids handed in.

The restriction needed a bind budget: at `per_page=100` the ids alone fill D1's cap, so the two window bounds moved inline the way `clickWindowSql` already inlines `sinceTs`. Above the cap the ids will not bind at all, and `listLinks` passing the whole catalog wants the full aggregate anyway, so that path keeps the unrestricted query. Both cases are tested, including delta correctness at 101 links.

src/pages/links.tsx
DennisAlund: Fixed in 4e5c81f. The enum is now `no-links | all-disabled | no-matches`, and the service picks between them from an unfiltered catalog count.

`?filter=disabled` on an all-active catalog now renders "No links match the current filter." The `all-disabled` copy also stopped naming the toggle: it points at the "Disabled" filter, in all three locales. `links.showDisabled` was dead in every locale with no `t()` call anywhere, so it is gone too.

One note on the page test: the layout embeds every translation string in the admin client script, so a page-wide `not.toContain` matches that blob rather than the markup. The assertions scope to the rendered `.empty-state` div.

src/services/link-management.ts
DennisAlund: Fixed in 4e5c81f. `emptyReason` keys off `result.links.length === 0` instead of `total === 0`, so the reason is set exactly when the page renders empty copy. A test asserts a reason is present whenever the served window is empty.

src/pages/links.tsx
DennisAlund: Fixed in bb587a6. `pageWindow(page, perPage, totalPages)` floors all three and the JSX reads only the clamped values, so `totalPages: 0` renders page 1 and a non-negative row range rather than "-24 - 0 of 0" and `page=0`.

The upper bound on `perPage` stays in the service, which owns `LINKS_MAX_PER_PAGE`; the component only floors it to 1. Six cases cover it, including NaN.

src/db/link-repository.ts
DennisAlund: Fixed in fee99f8. `page()` calls `LinkRepository.count(db, query)`. Statement count per render is unchanged at three, and a test walks seven filter combinations asserting `page().total` equals `count()`.

src/constants.ts
DennisAlund: Fixed in f89f208. `LINKS_MAX_PER_PAGE = Math.max(...LINKS_PER_PAGE_OPTIONS)`, with a test pinning the relation.

src/index.tsx
DennisAlund: Fixed in da298c2. Took the first option: `listLinksPage` returns `LinksPageData` directly. It clamped every input and had no `fail` branch, so the wrapper only bought an unreachable arm and a second copy of the page-data shape.

The service tests read the fields directly now that there is no wrapper to narrow.

src/services/index.ts
DennisAlund: Fixed in a9a6ccf. The page imports `LinksEmptyReason` through the barrel and the barrel carries that one type. `ListLinksOptions`, `ListLinksPageOptions` and `LinksPageData` stay exported from `link-management` for use inside the module and go back in the barrel when something outside asks.

src/db/filters.ts
DennisAlund: Fixed in 1936485. Both fragments append `clickWindowSql`, so the invariant is structural. A test pins the shared tail to one literal, since that pair counting the same clicks is what the whole popular sort rests on.

src/__tests__/repository/link-page.test.ts
DennisAlund: Fixed in 96168c5. The test pins `3` at both catalog sizes alongside the equality.

On the second point: `db.batch()` takes `D1PreparedStatement[]`, which can only come from `db.prepare()`, so a batch refactor stays counted through the existing trap. The path that would slip past is `db.exec()`, which takes raw SQL, so the trap covers that instead. The helper moved to `__tests__/setup.ts` as `spyDb` since it now records SQL for the query-shape tests too.

src/index.tsx
DennisAlund: Fixed in da298c2. The route parses `page` and `per_page` and clamps neither; `listLinksPage` owns every bound. The comment above the call says that accurately now. A service test drives 0, -3, 0.5 and NaN through it, and the existing route tests for the clamps are unchanged.

DennisAlund:
The two findings that could not be posted inline are fixed in 5d6d672 and aefcfee.

**`search()` fan-out and the two SQL definitions.** `LinkRepository.rows()` is now the unbounded sibling of `page()`. Both derive their predicate from `linkFilterSql`, so search, owner and status have one definition and one tie-break, and `list`, `search`, `findByOwner` and `findByUrl` all delegate to it. Each costs two statements whatever the match count: the rows, then the slugs of those rows bucketed by `link_id`. Tests assert two statements at 3 matches and two at 40, on all four methods.

`findByOwner` had the identical `Promise.all(ids.map(getById))` shape behind MCP `list_links_by_owner`, and `findByUrl` had it too, so all three went with the same change. `findByUrl`'s match count is bounded by duplicate URLs and was never the subrequest risk `search` was, but leaving one of four on the old shape is how the divergence comes back.

The slug query there restates the predicate as a subquery rather than binding ids: a match set has no ceiling, so ids cannot be bound the way `page()` binds its 100-row window. The subquery carries no ORDER BY and no LIMIT, so it costs a second pass over the predicate and no second sort.

**One behavior change on the API and MCP surfaces, and the index is what causes it.** I measured three links sharing a `created_at` second:

| | plan | tie order |
|---|---|---|
| pre-branch: `ORDER BY created_at DESC`, no index | `SCAN links \| USE TEMP B-TREE FOR ORDER BY` | `1,2,3` oldest first |
| same SQL, with `idx_links_created_at` | `SCAN links USING INDEX idx_links_created_at` | `3,2,1` newest first |
| `ORDER BY created_at DESC, id DESC` | index walk | `3,2,1` newest first |

So the migration flips tie ordering on its own. The explicit tie-break does not introduce the change, it pins it: without it the index would have flipped API and MCP ordering with nothing in the code saying so, and it would flip back silently the day the planner stops choosing that index.

Nothing depends on it. The spec makes exactly one ordering promise and it is about `BreakdownPage`. No SDK README or MCP tool description mentions link ordering. `GET /api/links` takes only `owner` and `range` with no limit/offset, so it returns whole arrays and unstable ties were never corrupting pagination, only producing nondeterministic order. A test now pins newest-first with the tie-break across `list`, `search` and `findByOwner`.

Result shapes are unchanged and `./scripts/spec-hash.sh` still matches all three recorded SDK hashes, so no spec drift and no SDK work.

---

Thirteen commits, one finding each. Full suite: 83 files, 1235 tests, all passing. `tsc --noEmit` clean.

Two things worth flagging:

- **`migrations/0007`** adds `idx_links_created_at`. Index only, no rows touched.
- **Two tests this branch added earlier changed.** The `all-filtered` service assertion renames to `all-disabled` because that value no longer exists, and the `link-page-service` tests read fields directly now that `listLinksPage` returns `LinksPageData` instead of a `ServiceResult`. No assertion was weakened; nothing pre-existing was touched.

I pushed back on half of one finding, on the `batch()` trap: `db.batch()` only accepts statements from `db.prepare()`, so a batch refactor stays counted. `db.exec()` is the path that would slip past, and the trap covers that instead.

PR #55, opened 2026-08-26
Name the search query in the links empty state

### LINKED ISSUE #52: A links search with no matches says "No links yet"
Found while reviewing #48. Pre-existing, not caused by that PR.

The links route sets `links` to the search results (`src/index.tsx:230`), so the empty-state branch at `src/pages/links.tsx:221` cannot tell an empty catalog from an empty result set.

## Reproduction

Catalog holds 200 links. Search for `zzz`.

- `links.length === 0`, so the branch falls to `t("links.empty")`: `No links yet. Use the + New Link button above to get started.`
- The user reads that their catalog is empty when it holds 200 links.

## Fix

Add a third empty-state branch keyed on `searchQuery`, with copy naming the query that matched nothing. Keys go in `src/i18n/en.ts`, `id.ts`, and `sv.ts`.

DennisAlund:
Closes #52.

## What was already fixed

The report says a search with no matches renders `No links yet. Use the + New Link button above to get started.` on a catalog holding 200 links. That symptom is gone on `main`: 4e5c81f (in #50) gave the service a three-value `emptyReason`, so an empty search result resolves to `no-matches` instead of falling through to the first-run copy.

## What was left

`no-matches` carries two different claims. Either a filter chip selected none of the catalog, or a search matched none of it. Both rendered the same string:

> No links match the current filter.

Searching `zzz` blamed a filter the user never touched, and never named the query that came back empty. That is the fix the issue asks for: "a third empty-state branch keyed on `searchQuery`, with copy naming the query that matched nothing."

## The change

The service hands back one reason and the page owns the query, so the split lands in the page. `emptyStateCopy` (`src/pages/links.tsx:124`) takes the reason and the query and picks `links.noSearchMatches` whenever a search is what emptied the window. Lifting the choice out of the JSX also keeps a fourth arm from growing the nested ternary another level.

New key in all three locales, per the CLAUDE.md i18n rule:

| Locale | `links.noSearchMatches` |
|---|---|
| en | `No links match "{query}".` |
| id | `Tidak ada tautan yang cocok dengan "{query}".` |
| sv | `Inga länkar matchar "{query}".` |

`Translations` is typed off `en`, so `id` and `sv` cannot lag behind the key.

The query reaches the copy as a `t()` parameter and hono/jsx escapes string children, so markup typed into the search box renders as text. A test pins that rather than trusting it.

## Tests

Three added to `src/__tests__/page/links-page.test.ts`, all watched failing first:

- `names the query in the empty state when a search matched nothing` asserts the query appears and neither `current filter` nor `No links yet` does.
- `keeps the filter wording when a filter, not a search, emptied the list` guards the filter case against the new branch swallowing it.
- `escapes markup in a search query before naming it in the empty state` asserts `&lt;b&gt;zzz&lt;/b&gt;` and no raw tag.

No existing test was modified or removed.

```
$ yarn tsc --noEmit
Done in 1.96s.

$ yarn test --run
Test Files  83 passed (83)
     Tests  1245 passed (1245)
```

No API surface change, so all three SDK spec hashes stay put: `./scripts/spec-hash.sh` still returns `d4cf1923…deb956`, matching what the manifests record.

### DIFF
--- a/src/__tests__/page/links-page.test.ts
+++ b/src/__tests__/page/links-page.test.ts
@@ -649,34 +649,30 @@ describe("Links listing page windowing", () => {
   });
 
   it("names the query in the empty state when a search matched nothing", async () => {
-    for (const slug of ["one", "two", "three"]) {
-      await LinkRepository.create(env.DB, { url: `https://example.com/${slug}`, slug });
-    }
+    await seed(3);
     const html = await (
       await SELF.fetch(req("/_/admin/links?search=zzz", { headers: { Cookie: "lang=en" } }))
     ).text();
 
     // The catalog is not empty and no filter chip emptied it: the search did.
-    expect(emptyState(html)).toContain("zzz");
+    // Match the whole sentence, so a bare query or an unreplaced placeholder
+    // sitting next to one cannot pass.
+    expect(emptyState(html)).toContain("No links match &quot;zzz&quot;.");
     expect(emptyState(html)).not.toContain("current filter");
     expect(emptyState(html)).not.toContain("No links yet");
   });
 
   it("keeps the filter wording when a filter, not a search, emptied the list", async () => {
-    for (const slug of ["one", "two", "three"]) {
-      await LinkRepository.create(env.DB, { url: `https://example.com/${slug}`, slug });
-    }
+    await seed(3);
     const html = await (
       await SELF.fetch(req("/_/admin/links?filter=disabled", { headers: { Cookie: "lang=en" } }))
     ).text();
 
-    expect(emptyState(html)).toContain("current filter");
+    expect(emptyState(html)).toContain("No links match the current filter.");
   });
 
   it("escapes markup in a search query before naming it in the empty state", async () => {
-    for (const slug of ["one", "two", "three"]) {
-      await LinkRepository.create(env.DB, { url: `https://example.com/${slug}`, slug });
-    }
+    await seed(3);
     const html = await (
       await SELF.fetch(
         req(`/_/admin/links?search=${encodeURIComponent("<b>zzz</b>")}`, {
@@ -687,6 +683,19 @@ describe("Links listing page windowing", () => {
 
     expect(emptyState(html)).toContain("&lt;b&gt;zzz&lt;/b&gt;");
     expect(emptyState(html)).not.toContain("<b>zzz</b>");
+  });
+
+  it("treats a whitespace-only search as no search at all", async () => {
+    // The repository returns nothing for a search that trims to empty, so an
+    // untrimmed query reaches the empty state and gets attributed to whatever
+    // the filter happens to be. Under the default Active filter that reads
+    // "All links are disabled" over a catalog where nothing is disabled.
+    await seed(3);
+    for (const url of ["/_/admin/links?search=%20%20", "/_/admin/links?filter=all&search=%20"]) {
+      const html = await (await SELF.fetch(req(url, { headers: { Cookie: "lang=en" } }))).text();
+      expect(emptyState(html)).toBe("");
+      expect(rowCount(html)).toBe(3);
+    }
   });
 
   it("does not print a row count above an empty state", async () => {
--- a/src/__tests__/page/links-page.test.ts
+++ b/src/__tests__/page/links-page.test.ts
@@ -654,12 +654,40 @@ describe("Links listing page windowing", () => {
       await SELF.fetch(req("/_/admin/links?search=zzz", { headers: { Cookie: "lang=en" } }))
     ).text();
 
-    // The catalog is not empty and no filter chip emptied it: the search did.
-    // Match the whole sentence, so a bare query or an unreplaced placeholder
-    // sitting next to one cannot pass.
-    expect(emptyState(html)).toContain("No links match &quot;zzz&quot;.");
+    // The search emptied the window, and the default Active chip is narrowing
+    // too, so the copy names both. Match the whole sentence, so a bare query or
+    // an unreplaced placeholder sitting next to one cannot pass.
+    expect(emptyState(html)).toContain("No links match &quot;zzz&quot; in Active.");
     expect(emptyState(html)).not.toContain("current filter");
     expect(emptyState(html)).not.toContain("No links yet");
+  });
+
+  it("names the filter alongside the query when a chip is narrowing too", async () => {
+    // A link matching "one" exists and is one chip away, so blaming the query
+    // alone would be as wrong as blaming the filter alone.
+    for (const slug of ["one", "two", "three"]) {
+      await LinkRepository.create(env.DB, { url: `https://example.com/${slug}`, slug });
+    }
+    const html = await (
+      await SELF.fetch(
+        req("/_/admin/links?filter=disabled&search=one", { headers: { Cookie: "lang=en" } }),
+      )
+    ).text();
+
+    expect(emptyState(html)).toContain("No links match &quot;one&quot; in Disabled.");
+  });
+
+  it("names only the query when the All chip is hiding nothing", async () => {
+    await seed(3);
+    const html = await (
+      await SELF.fetch(
+        req("/_/admin/links?filter=all&search=zzz", { headers: { Cookie: "lang=en" } }),
+      )
+    ).text();
+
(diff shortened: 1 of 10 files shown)


DennisAlund:
Self-review pass over this branch. Ten findings; seven applied in 016a36a, two need a call from you, one I disagree with.

## Applied

**Whitespace-only search rendered a false claim.** `?search=%20` on a full catalog: the repository treats a query that trims to empty as matching nothing, rather than letting a bare `LIKE "%%"` match every row, so the window comes back empty and the reason gets picked from the filter alone. Under the default Active chip that read `All links are disabled. Pick the "Disabled" filter to see them.` over a catalog where nothing is disabled. Under `?filter=all` it read `No links match the current filter.` with the filter set to All. Both are pre-existing on `main`, and both are exactly the class of bug #52 is about, so they belong here. The route trims the query before anything reads it, so a run of spaces is no search and the listing just renders.

**Non-exhaustive reason test.** `emptyReason !== "no-matches"` mapped anything unrecognised onto the first-run copy, so a fourth `LinksEmptyReason` would have shipped `No links yet` over a full catalog with no build error. Now an exhaustive `switch` with a `never`-typed default. Confirmed rather than assumed: adding a `"probe"` member to the union gives `src/pages/links.tsx(162,13): error TS2322: Type '"probe"' is not assignable to type 'never'`.

**No placeholder parity guard.** The PR body claimed `Translations` being typed off `en` keeps the locales honest. That is true for key presence and false for the `{query}` inside a value: a locale could keep the key and drop the interpolation it exists for. New test in `i18n.test.ts`. Confirmed it bites: dropping `{query}` from `id.ts` fails it. Every existing key already passes, so it locks in current state.

**Unbounded query length.** The query now clips at 60 characters with an ellipsis. Escaping already made it safe; this stops a pasted essay from filling the viewport and pushing the toolbar and paginator out of view.

**Helper not exported.** `emptyStateCopy` now sits next to `pageWindow` and `paginationItems` as an exported pure function, with the branch matrix tabled in `src/__tests__/unit/links-empty-state.test.ts`. The whitespace and long-query cases cost a function call there instead of a DB seed plus a `SELF.fetch`.

**Weak headline assertion.** `toContain("zzz")` passed on a bare query or an unreplaced placeholder sitting next to one. Now asserts `No links match &quot;zzz&quot;.`

**Duplicated seeding.** The three page tests use the `seed(3)` helper the enclosing describe already provides.

## Needs your call

**1. Search plus a narrowing filter blames only the query.** `?filter=disabled&search=one` with three active links slugged one/two/three renders `No links match "one".` A link matching `one` does exist, one chip away. This is a genuine regression from the first commit: `main` renders `No links match the current filter.` there, which is the more accurate half of the pair. There is no short sentence that is honest in both directions, so it needs a copy decision, and every option collides with an assertion already on this branch. Options: name the filter too (`No links match "one" in Disabled.`), always say `under the current filter`, or accept the current wording as scoped to the visible view. I did not want to pick UX copy for three locales unilaterally.

**2. The split arguably belongs in the service.** `listLinksPage` already computes `search.trim()` to choose `all-disabled` vs `no-matches`, and the page recomputes it. A fourth reason `no-search-matches` emitted where the trim already lives would give one source of truth and keep the page a pure reason-to-string map. I did not do it because it breaks `reports no-matches when a search finds nothing in a populated catalog` (`link-page-service.test.ts:88`), and CLAUDE.md says to stop and notify rather than modify an existing test. Route-level trimming removes the divergence in practice, so this is now structure rather than a live bug. Say the word and I will do it as a follow-up with the test change made deliberately.

## Disagree

**Fold the filter-regression test into the existing one.** Adding the `current filter` assertion to `does not claim every link is disabled when the disabled filter matched nothing` would edit a committed test, which CLAUDE.md forbids outright. A separate test is the compliant shape. It does now share the `seed(3)` helper, so the duplication is four lines rather than ten.

## Verification

```
$ yarn tsc --noEmit
Done in 2.13s.

$ yarn test --run
Test Files  84 passed (84)
     Tests  1255 passed (1255)
```

The `uncaught exception; internal error` lines in that output are workers-pool noise, not this branch: five occurrences on `main` and five here.

DennisAlund:
Both open decisions are settled and pushed in c8fbe92.

## Copy names both narrowing dimensions

| Request | Empty state |
|---|---|
| `?search=zzz` (Active is the default chip) | `No links match "zzz" in Active.` |
| `?filter=disabled&search=one` | `No links match "one" in Disabled.` |
| `?filter=all&search=zzz` | `No links match "zzz".` |
| `?filter=disabled` (no search) | `No links match the current filter.` |

Under `all` nothing but the search is hiding rows, so there is no filter worth naming. Under `active` or `disabled` both are narrowing, and blaming either alone is half the truth: the case that prompted this was `?filter=disabled&search=one` rendering `No links match "one".` with a matching link sitting one chip away.

`links.noSearchMatchesInFilter` added to all three locales. The chip table moved to module scope as `FILTER_CHIPS`, so the chip a user clicks and the filter the copy names read from one place and cannot drift.

## Split moved into the service

`LinksEmptyReason` gains `no-search-matches`, raised by `listLinksPage` wherever the trimmed query is what emptied the window. The page no longer re-derives "did a search happen" from the raw query one layer above the trim that already answered it. A live search outranks the status filter when picking the reason, since the query is the specific thing the copy can name back. Every other reason is unchanged.

This renamed the assertion in `reports no-matches when a search finds nothing in a populated catalog` (`link-page-service.test.ts:88`). CLAUDE.md forbids editing an existing test to fit a code change, so that conflict went to Dennis before anything was touched, and the rename is deliberate: the value it named no longer describes that case. A second service test covers a search and a status filter narrowing together.

## Verification

```
$ yarn tsc --noEmit
Done in 2.10s.

$ yarn test --run
Test Files  84 passed (84)
     Tests  1259 passed (1259)
```

The exhaustiveness guard was re-checked after the union grew: adding a `"probe"` member still gives `src/pages/links.tsx(180,13): error TS2322: Type '"probe"' is not assignable to type 'never'`. The placeholder parity test covers the new two-placeholder key across `id` and `sv`.

`./scripts/spec-hash.sh` still returns `d4cf1923…deb956`, so no SDK hash bumps: `LinksEmptyReason` is admin-page plumbing and never reaches the OpenAPI surface.

DennisAlund:
Fresh read of the whole branch, treating it as someone else's work. Two findings, both fixed in 0ae8233. Two more checked and dismissed.

## Fixed

**The clip cut UTF-16 units, not characters.** A query long enough to trim, with its emoji on odd offsets, lost half a surrogate pair at the boundary and shipped an unpaired surrogate that renders as U+FFFD. `[...query]` counts and cuts code points instead.

Worth recording how this nearly slipped through: the first version of the test used an all-emoji query, where every pair straddles the 60th unit evenly and a naive slice gets away with it. It passed against the bug. A leading letter shifts every pair onto an odd offset and the cut lands inside one:

```
"👍".repeat(40).slice(0,60)        -> lone surrogate: false
("a"+"👍".repeat(40)).slice(0,60)  -> lone surrogate: true
```

The corrected test was run against the old slice to confirm it fails there before the fix went in.

**The filter label had an unreachable arm.** `FILTER_CHIPS.find((chip) => chip.key === filter)?.labelKey` returns `undefined` for a key not in the list, forcing `label ? t(label) : filter`. `LinksFilter` has three members and the list covers all three, so that fallback could never run and existed only to satisfy the type. `FILTER_LABEL` is a `Record<LinksFilter, TranslationKey>`, total by construction: no fallback, and a filter added to the union now fails the build here instead of silently rendering its raw key. The chips read their labels from the same table.

## Checked and dismissed

**Adding a member to `LinksEmptyReason` could strand a consumer.** Grepped: `listLinksPage` and `LinksPage` have exactly one caller each, both in `src/index.tsx`. No widget island reads the reason yet, so nothing can receive `no-search-matches` without handling it.

**A search over a fully expired catalog now reports the search rather than `all-disabled`.** Real, and deliberate. `?search=one` with every link expired renders `No links match "one" in Active.`, which is true: Active holds nothing. `main` rendered `No links match the current filter.` there, so this is not a regression, and the query is the more specific thing to name. Left as is.

## Verification

```
$ yarn tsc --noEmit
Done in 2.09s.

$ yarn test --run
Test Files  84 passed (84)
     Tests  1260 passed (1260)
```

PR #59, opened 2026-09-11
Weekly review batch: silent custom-slug failure, unsafe client-side interpolation, Python SDK InvalidURL escape, slug-repository invariant guard, unscannable QR codes, title-fetch SSRF guard

DennisAlund:
Weekly defect-hunting review of the app, its API, and all three SDKs. Six parallel passes (API/DB/services, admin UI, MCP/auth, TypeScript SDK, Python SDK, Dart SDK), each finding verified against the actual code and a failing-before/passing-after regression test before being fixed.

## Fixes

**`createLink()`'s custom-slug attach toasted "Link created" whether it succeeded or failed** (`src/client.ts`)
- Wrong: the "New Link" modal's custom-slug branch had `if (!slugRes.ok) { toast(t('client.linkCreated')); } else { toast(t('client.linkCreated')); }` — both branches identical.
- Impact: requesting a custom slug that's already taken (or otherwise rejected) failed silently. The user was told "Link created" and redirected to the link's detail page with no sign the slug they asked for never attached.
- Fix: extracted the step into `attachCustomSlugAndGo()`, which reads the failure body and toasts `client.customError` (or the server's message), matching `doAddSlug`'s existing handling of the same endpoint.
- Test: drives `createLink()` against a stubbed `api()` that succeeds on the link-create call and fails on the slug-attach call; asserts the failure surfaces, the generic fallback fires when the body carries no message, and the success toast still fires when the slug attaches.

**Client-side `t()` used unsafe string-based interpolation** (`src/client.ts`)
- Wrong: `val.replace(new RegExp(...), String(params[k]))` — the plain-string form of `.replace()` treats `$&`, `` $` ``, `$'`, `$1` etc. in the *replacement string* as special patterns instead of literal text. The server-side `createTranslateFn()` in `src/i18n/index.ts` already guards against exactly this with a replacer function; the client-side reimplementation didn't.
- Impact: any `$`-containing free text passed as a param corrupts the shown string. Two real call sites: API key titles (`client.confirmDeleteKey`) and bundle names (`client.bundles.confirmArchive`/`confirmDelete`), both unrestricted free text. A title like `Prod $$ key` renders as `Prod $ key`.
- Fix: pass a replacer function, mirroring the server-side fix.
- Test: extracts `t()` from the generated script and asserts literal interpolation, mirroring the existing server-side regression test for the same bug class.

**Python SDK: `httpx.InvalidURL` escaped as a raw exception instead of `ShrtnrError`** (`sdk/python/src/shrtnr/resources/{links,bundles,slugs}.py`)
- Wrong: every resource's `_request`/`_request_text` (8 call sites, sync and async) caught only `httpx.RequestError`. `httpx.InvalidURL` (e.g. an unparsable port) is not a subclass of it, and is raised synchronously from request-building, before any I/O.
- Impact: a malformed `base_url` (bad port, bad IDNA host, etc.) raised a raw `httpx.InvalidURL` instead of the documented `ShrtnrError(status=0, ...)`, breaking the "network failures wrap as `ShrtnrError`" contract in `errors.py` and the README.
- Fix: widened the caught exception tuple to `(httpx.RequestError, httpx.InvalidURL)` at all 8 sites.
- Test: one regression test per sync/async client, constructing a client with a malformed `base_url` and asserting `ShrtnrError(status=0)`.
- Single-SDK change: this is an httpx-specific exception-hierarchy quirk (`InvalidURL` isn't related to `RequestError`). The TypeScript SDK's fetch-based error handling and the Dart SDK's `http` client don't have an equivalent split, so there's nothing to port.

**`SlugRepository.disable()` could zero out a link's primary flag** (`src/db/slug-repository.ts`)
- Wrong: a link has exactly one system-generated (`is_custom = 0`) slug for its whole lifetime. `disable()`'s primary-fallback query promotes by `is_custom = 0`, so disabling that one row has the query re-select the very row being demoted next, leaving the link with no primary slug at all.
- Impact: currently unreachable — the only caller (`disableSlug` in `link-management.ts`) already refuses to disable a non-custom slug before reaching the repository — but a landmine for any future caller (MCP tool, script, admin action) that calls the repository directly.
- Fix: added the same guard `SlugRepository.remove()` already has, so the repository enforces its own invariant rather than relying solely on the one caller.
- Test: calls `SlugRepository.disable()` directly on a link's only slug (bypassing the service-layer guard) and asserts it refuses, leaving the primary flag untouched.

**QR encoder produced unscannable codes for versions 6-10** (`src/qr.ts`)
- Wrong: three stacked defects, each confirmed with a real decoder (jsQR). (1) Versions 6-10 must split codewords into 2 (v6-9) or 4 (v10) Reed-Solomon blocks and interleave them; the encoder read the per-block ECC count as the total and ran one RS pass over the whole payload. (2) Versions 7-10 must carry two 18-bit BCH-protected version-information blocks; the encoder never reserved or wrote them, so data modules landed where a reader looks for the version. (3) From version 7 on, alignment patterns also sit on the timing pattern's row and column; the "skip if reserved" test meant for the three finder overlaps also skipped those.
- Impact: any QR payload over 106 bytes (a short link with a long custom slug plus `?utm_medium=qr` gets there) rendered a symbol no reader could decode. Versions 1-5 were unaffected.
- Fix: per-version block-count table with proper block splitting and interleaving, version-information blocks reserved and written for v7+, and an index-based finder-overlap skip for alignment patterns.
- Test: rasterizes each version's output and decodes it with jsQR (new dev dependency), asserts the version-info bits against ISO/IEC 18004 Table D.1, and checks the 271-byte capacity boundary. Before the fix versions 6-10 failed to decode; after it every payload length from 1 to 271 bytes round-trips.

**`fetchPageTitle` fetched any user-supplied URL server-side with no restriction** (`src/title-fetch.ts`)
- Wrong: `autoLabelLink()` fetched the link URL with `redirect: "follow"` and no host check, an SSRF pattern. A link at a loopback, RFC 1918, link-local or cloud-metadata address (or a public page redirecting to one) had the Worker fetch it and store the response's `<title>` where the link's owner can read it.
- Impact: bounded in practice by Cloudflare Workers' own egress isolation, but the code assumed that instead of enforcing it.
- Fix: new `isPublicHttpUrl()` admits only http(s) URLs whose host is a public name or public IP literal, rejecting localhost/`.localhost`/`.internal`/`.local` names and every reserved IPv4 and IPv6 range (IPv4-mapped and NAT64 literals judged by the embedded IPv4 address; alternate IPv4 spellings are normalized by the URL parser first). `fetchPageTitle()` fetches with `redirect: "manual"` and follows up to five hops itself, vetting each `Location` before requesting it. Hostnames are not resolved (Workers expose no DNS), so a public name pointing at a private address is left to the platform.
- Test: 27 blocked targets across schemes, names and IPv4/IPv6 ranges assert `fetch` is never called; redirect tests cover public-to-public, relative `Location`, redirect-to-private, redirect-to-`file:`, redirect loops and a missing `Location`.

## Verification
- `yarn test`: 1360 tests, all green.
- `yarn e2e`: 15 failures reproduce identically (same test names, same count) on a clean `main` checkout in this sandbox — `net::ERR_CONNECTION_RESET`/browser-launch issues from the sandbox's network policy, unrelated to this diff. Verified via a side-by-side worktree run against `origin/main`.
- Python SDK: 104 tests, all green. `ruff check`/`ruff format --check`/`mypy` clean on changed files.

---

https://claude.ai/code/session_017a43SYLJdvdDhMcgJF4zXM

### DIFF
--- a/sdk/python/src/shrtnr/resources/links.py
+++ b/sdk/python/src/shrtnr/resources/links.py
@@ -52,16 +52,22 @@ class Links:
         return _build_url(self._base_url, path, query)
 
     def _request(self, method: str, url: str, **kwargs: Any) -> Any:
+        # httpx.InvalidURL isn't a RequestError subclass; catch it too so a
+        # malformed base_url raises ShrtnrError(0, ...) like any other request
+        # failure, per this SDK's documented contract.
         try:
             response = self._http.request(method, url, **kwargs)
-        except httpx.RequestError as exc:
+        except (httpx.RequestError, httpx.InvalidURL) as exc:
             raise ShrtnrError(0, str(exc)) from exc
         return parse_json_response(response)
 
     def _request_text(self, method: str, url: str, **kwargs: Any) -> str:
+        # httpx.InvalidURL isn't a RequestError subclass; catch it too so a
+        # malformed base_url raises ShrtnrError(0, ...) like any other request
+        # failure, per this SDK's documented contract.
         try:
             response = self._http.request(method, url, **kwargs)
-        except httpx.RequestError as exc:
+        except (httpx.RequestError, httpx.InvalidURL) as exc:
             raise ShrtnrError(0, str(exc)) from exc
         return parse_text_response(response)
 
@@ -206,16 +212,22 @@ class AsyncLinks:
         return _build_url(self._base_url, path, query)
 
     async def _request(self, method: str, url: str, **kwargs: Any) -> Any:
+        # httpx.InvalidURL isn't a RequestError subclass; catch it too so a
+        # malformed base_url raises ShrtnrError(0, ...) like any other request
+        # failure, per this SDK's documented contract.
         try:
             response = await self._http.request(method, url, **kwargs)
-        except httpx.RequestError as exc:
+        except (httpx.RequestError, httpx.InvalidURL) as exc:
             raise ShrtnrError(0, str(exc)) from exc
         return parse_json_response(response)
 
     async def _request_text(self, method: str, url: str, **kwargs: Any) -> str:
+        # httpx.InvalidURL isn't a RequestError subclass; catch it too so a
+        # malformed base_url raises ShrtnrError(0, ...) like any other request
+        # failure, per this SDK's documented contract.
         try:
             response = await self._http.request(method, url, **kwargs)
-        except httpx.RequestError as exc:
+        except (httpx.RequestError, httpx.InvalidURL) as exc:
             raise ShrtnrError(0, str(exc)) from exc
         return parse_text_response(response)
 
--- a/sdk/python/tests/test_async_client.py
+++ b/sdk/python/tests/test_async_client.py
@@ -92,6 +92,17 @@ async def test_async_network_error_wraps_as_status_0(client: AsyncShrtnr) -> Non
         assert exc_info.value.status == 0
 
 
+async def test_async_malformed_base_url_wraps_as_status_0() -> None:
+    # httpx.InvalidURL (e.g. a bad port) is raised synchronously from
+    # request-building, before any I/O, and is not a subclass of
+    # httpx.RequestError. It must still surface as the documented
+    # ShrtnrError(status=0, ...), not escape as a raw httpx exception.
+    client = AsyncShrtnr(base_url="https://example.com:notaport", api_key=API_KEY)
+    with pytest.raises(ShrtnrError) as exc_info:
+        await client.links.list()
+    assert exc_info.value.status == 0
+
+
 # ---- Links resource ----
 
 
--- a/src/__tests__/unit/client-custom-slug-toast.test.ts
+++ b/src/__tests__/unit/client-custom-slug-toast.test.ts
@@ -1,0 +1,136 @@
+// Copyright 2026 Oddbit (https://oddbit.id)
+// SPDX-License-Identifier: Apache-2.0
+
+// createLink() makes two API calls when the "New Link" modal's custom-slug
+// field is filled in: create the link, then attach the custom slug. The
+// second call's success and failure branches used to show the identical
+// "Link created" toast either way, so a rejected custom slug (already
+// taken, invalid) failed silently: the user was told the slug they asked
+// for was attached when it wasn't. This drives createLink() against a
+// stubbed api() that succeeds on the first call and fails on the second,
+// and asserts the failure actually surfaces.
+import { describe, expect, it } from "vitest";
+import { adminClientScript } from "../../client";
+import type { Translations } from "../../i18n/types";
+
+function extractTopLevelChunk(source: string, startPattern: RegExp): string {
+  const lines = source.split("\n");
+  const startIdx = lines.findIndex((l) => startPattern.test(l));
+  if (startIdx === -1) throw new Error(`chunk not found: ${startPattern}`);
+  let endIdx = lines.length;
+  for (let i = startIdx + 1; i < lines.length; i++) {
+    if (/^(function |var |if |window\.|document\.)/.test(lines[i])) {
+      endIdx = i;
+      break;
+    }
+  }
+  return lines.slice(startIdx, endIdx).join("\n");
+}
+
+type ToastCall = { message: string; level?: string };
+
+// Fields read by createLink(): url and custom are the interesting ones,
+// label/len/expires are left blank so the request body stays minimal.
+function fakeDocument() {
+  const values: Record<string, string> = {
+    "m-url": "https://example.com",
+    "m-label": "",
+    "m-len": "",
+    "m-custom": "taken-slug",
+    "m-expires": "",
+  };
+  return {
+    getElementById: (id: string) => ({ value: values[id] ?? "", focus() {} }),
+  };
+}
+
+// First api() call (create the link) resolves ok with a link id; second
+// call (attach the custom slug) resolves with whatever `slugOk`/`slugJson`
+// says, mirroring a real POST /links/:id/slugs response.
+function loadCreateLink(slugOk: boolean, slugJson: () => Promise<unknown>) {
+  const script = adminClientScript("1.0.0", {} as unknown as Translations);
+  const code = [
+    extractTopLevelChunk(script, /^function createLink\(/),
+    extractTopLevelChunk(script, /^function attachCustomSlugAndGo\(/),
+    "return { createLink: createLink };",
+  ].join("\n");
+
+  const toasts: ToastCall[] = [];
+  let call = 0;
+  const api = () => {
+    call += 1;
+    if (call === 1) {
+      return Promise.resolve({
+        ok: true,
+        status: 201,
+        json: () => Promise.resolve({ id: 42 }),
+      });
+    }
+    return Promise.resolve({ ok: slugOk, status: slugOk ? 200 : 409, json: slugJson });
+  };
+
+  // eslint-disable-next-line @typescript-eslint/no-implied-eval
+  const factory = new Function(
+    "api",
+    "toast",
+    "t",
+    "closeModal",
+    "window",
+    "document",
+    code,
+  ) as (...args: unknown[]) => { createLink: () => void };
+
+  const window_ = { location: { href: "" } };
+  const handlers = factory(
+    api,
+    (message: string, level?: string) => {
+      toasts.push({ message, level });
+    },
+    (key: string) => key,
+    () => {},
+    window_,
+    fakeDocument(),
+  );
+  return { handlers, toasts, window: window_ };
+}
+
+async function waitForToast(toasts: ToastCall[]) {
+  for (let i = 0; i < 50 && toasts.length === 0; i++) {
+    await new Promise((resolve) => setTimeout(resolve, 0));
+  }
+}
+
+describe("createLink() custom-slug attach toast", () => {
+  it("toasts a distinct error when attaching the requested custom slug fails", async () => {
+    const { handlers, toasts, window: win } = loadCreateLink(false, () =>
+      Promise.resolve({ error: "slug already taken" }),
+    );
+
+    handlers.createLink();
+    await waitForToast(toasts);
+
+    expect(toasts).toEqual([{ message: "slug already taken", level: "error" }]);
+    // The link itself was created, so the user still lands on its detail page.
+    expect(win.location.href).toBe("/_/admin/links/42");
+  });
+
+  it("falls back to a generic custom-slug error when the failure body has no message", async () => {
+    const { handlers, toasts } = loadCreateLink(false, () => Promise.reject(new Error("no body")));
+
+    handlers.createLink();
+    await waitForToast(toasts);
+
+    expect(toasts).toHaveLength(1);
+    expect(toasts[0].level).toBe("error");
+    expect(toasts[0].message).toBe("client.customError");
+  });
+
+  it("toasts the success message when the custom slug attaches", async () => {
+    const { handlers, toasts } = loadCreateLink(true, () => Promise.resolve({}));
+
+    handlers.createLink();
+    await waitForToast(toasts);
+
+    expect(toasts).toEqual([{ message: "client.linkCreated", level: undefined }]);
+  });
+});
--- a/src/__tests__/unit/client-t-replacer.test.ts
+++ b/src/__tests__/unit/client-t-replacer.test.ts
@@ -1,0 +1,60 @@
+// Copyright 2026 Oddbit (https://oddbit.id)
+// SPDX-License-Identifier: Apache-2.0
+
+// The admin client script's t() interpolates {param} placeholders into a
+// translation string. src/i18n/index.ts's createTranslateFn() already
+// guards this with a replacer function instead of a plain string, because
+// String.prototype.replace treats "$&", "$`", "$'", "$1" etc. in a string
+// replacement as special patterns instead of literal text. t() in
+// client.ts reimplements the same interpolation client-side and needs the
+// same guard, or a param value containing one of those sequences (an API
+// key title, a bundle name — both free text) corrupts the rendered string.
+import { describe, expect, it } from "vitest";
+import { adminClientScript } from "../../client";
+import type { Translations } from "../../i18n/types";
+
+function extractTopLevelChunk(source: string, startPattern: RegExp): string {
+  const lines = source.split("\n");
+  const startIdx = lines.findIndex((l) => startPattern.test(l));
+  if (startIdx === -1) throw new Error(`chunk not found: ${startPattern}`);
+  let endIdx = lines.length;
+  for (let i = startIdx + 1; i < lines.length; i++) {
+    if (/^(function |var |if |window\.|document\.)/.test(lines[i])) {
+      endIdx = i;
+      break;
+    }
+  }
+  return lines.slice(startIdx, endIdx).join("\n");
+}
+
+function loadT(translations: Record<string, string>) {
+  const script = adminClientScript("1.0.0", {} as unknown as Translations);
+  const code = [
+    `var T = ${JSON.stringify(translations)};`,
+    extractTopLevelChunk(script, /^function t\(/),
+    "return t;",
+  ].join("\n");
+  // eslint-disable-next-line @typescript-eslint/no-implied-eval
+  const factory = new Function(code) as () => (key: string, params?: Record<string, unknown>) => string;
+  return factory();
+}
+
+describe("client.ts t()", () => {
+  it("interpolates a plain {param} value", () => {
+    const t = loadT({ greeting: "Hello {name}!" });
+    expect(t("greeting", { name: "World" })).toBe("Hello World!");
+  });
+
+  it("treats interpolated values as literal text, not replacement patterns", () => {
+    const t = loadT({ confirm: 'Delete "{title}"?' });
+    // "$`" is a special String.replace pattern (text before the match). A
+    // naive `val.replace(re, String(v))` would splice that in instead of
+    // the literal value.
+    expect(t("confirm", { title: "x$`y" })).toBe('Delete "x$`y"?');
+  });
+
+  it("does not collapse a literal $$ in the interpolated value", () => {
+    const t = loadT({ label: "{name}" });
+    expect(t("label", { name: "Q1 $$" })).toBe("Q1 $$");
+  });
+});
--- a/src/__tests__/unit/qr.test.ts
+++ b/src/__tests__/unit/qr.test.ts
@@ -72,3 +72,94 @@ describe("makeQR", () => {
     expect(makeQR(tooLong)).toBeNull();
   });
 });
+
+// Scannability: a real decoder (jsQR) must read back what makeQR() wrote.
+// The rendering assertions above cannot see a broken error-correction
+// layout or a missing version-info block; a decoder can.
+import jsQR from "jsqr";
+
+function rasterize(matrix: boolean[][], scale = 4, quiet = 4): { data: Uint8ClampedArray; width: number } {
+  const modules = matrix.length + quiet * 2;
+  const width = modules * scale;
+  const data = new Uint8ClampedArray(width * width * 4).fill(255);
+  for (let r = 0; r < matrix.length; r++) {
+    for (let c = 0; c < matrix.length; c++) {
+      if (!matrix[r][c]) continue;
+      for (let dy = 0; dy < scale; dy++) {
+        for (let dx = 0; dx < scale; dx++) {
+          const x = (c + quiet) * scale + dx;
+          const y = (r + quiet) * scale + dy;
+          const i = (y * width + x) * 4;
+          data[i] = 0;
+          data[i + 1] = 0;
+          data[i + 2] = 0;
+        }
+      }
+    }
+  }
+  return { data, width };
+}
+
+function decode(matrix: boolean[][]): string | null {
+  const { data, width } = rasterize(matrix);
+  return jsQR(data, width, width)?.data ?? null;
+}
+
+// Byte-mode capacity at error-correction level L, indexed by version.
+const CAPACITY_L = [0, 17, 32, 53, 78, 106, 134, 154, 192, 230, 271];
+
+describe("makeQR scannability", () => {
+  for (let ver = 1; ver <= 10; ver++) {
+    it(`version ${ver}: a payload filling the version decodes back to the same text`, () => {
+      const prefix = "https://s.ex/";
+      const text = prefix + "a".repeat(CAPACITY_L[ver] - prefix.length);
+      const grid = makeQR(text);
+      expect(grid).not.toBeNull();
+      expect(grid!.length).toBe(ver * 4 + 17);
+      expect(decode(grid!)).toBe(text);
+    });
+  }
+
+  it("decodes a realistic short link with utm_medium=qr", () => {
+    const text = "https://s.example/very-long-custom-slug-for-a-marketing-campaign?utm_medium=qr";
+    expect(decode(makeQR(text)!)).toBe(text);
+  });
+
+  it("encodes exactly 271 bytes (version 10 capacity) and refuses 272", () => {
+    expect(makeQR("x".repeat(271))).not.toBeNull();
+    expect(makeQR("x".repeat(272))).toBeNull();
+  });
+});
+
+describe("makeQR version information", () => {
+  // 18-bit version information (6 data bits + 12 BCH bits), ISO/IEC 18004 Table D.1.
+  const VERSION_INFO: Record<number, number> = {
+    7: 0x07c94,
+    8: 0x085bc,
+    9: 0x09a99,
+    10: 0x0a4d3,
+  };
+
+  function readVersionInfo(grid: boolean[][]): { topRight: number; bottomLeft: number } {
+    const size = grid.length;
+    let topRight = 0;
+    let bottomLeft = 0;
+    for (let i = 0; i < 18; i++) {
+      const a = size - 11 + (i % 3);
+      const b = Math.floor(i / 3);
+      if (grid[b][a]) topRight |= 1 << i;
+      if (grid[a][b]) bottomLeft |= 1 << i;
+    }
+    return { topRight, bottomLeft };
+  }
+
+  for (const ver of [7, 8, 9, 10]) {
+    it(`version ${ver}: both version-info blocks carry the BCH-encoded version`, () => {
+      const grid = makeQR("x".repeat(CAPACITY_L[ver - 1] + 1))!;
+      expect(grid.length).toBe(ver * 4 + 17);
+      const { topRight, bottomLeft } = readVersionInfo(grid);
+      expect(topRight).toBe(VERSION_INFO[ver]);
+      expect(bottomLeft).toBe(VERSION_INFO[ver]);
+    });
+  }
+});
--- a/src/client.ts
+++ b/src/client.ts
@@ -34,7 +34,10 @@ function t(key, params) {
   var val = T[key] || key;
   if (params) {
     for (var k in params) {
-      val = val.replace(new RegExp('\\\\{' + k + '\\\\}', 'g'), String(params[k]));
+      // Replacer function, not a string: a string replacement treats
+      // "$&", "$\`", "$'", "$1" etc. in the value as special patterns
+      // instead of literal text.
+      val = val.replace(new RegExp('\\\\{' + k + '\\\\}', 'g'), function() { return String(params[k]); });
     }
   }
   return val;
--- a/src/client.ts
+++ b/src/client.ts
@@ -257,21 +257,31 @@ function createLink() {
           window.location.href = '/_/admin/links/' + link.id;
           return;
         }
-        api('/links/' + link.id + '/slugs', { method: 'POST', body: JSON.stringify({ slug: custom }) }).then(function(slugRes) {
-          closeModal();
-          if (!slugRes.ok) {
-            toast(t('client.linkCreated'));
-          } else {
-            toast(t('client.linkCreated'));
-          }
-          window.location.href = '/_/admin/links/' + link.id;
-        });
+        attachCustomSlugAndGo(link.id, custom);
       });
     } else {
       return res.json().then(function(data) {
         toast(data.error || t('client.createLinkError'), 'error');
       }).catch(function() { toast(t('client.createLinkError'), 'error'); });
     }
+  });
+}
+
+// Attaches a custom slug to a just-created link, reporting whether the
+// slug itself was accepted (it may already be taken) separately from the
+// link creation that already succeeded by the time this runs.
+function attachCustomSlugAndGo(linkId, slug) {
+  api('/links/' + linkId + '/slugs', { method: 'POST', body: JSON.stringify({ slug: slug }) }).then(function(slugRes) {
+    closeModal();
+    if (!slugRes.ok) {
+      return slugRes.json().then(function(data) {
+        toast(data.error || t('client.customError'), 'error');
+      }).catch(function() { toast(t('client.customError'), 'error'); }).then(function() {
+        window.location.href = '/_/admin/links/' + linkId;
+      });
+    }
+    toast(t('client.linkCreated'));
+    window.location.href = '/_/admin/links/' + linkId;
   });
 }
 
--- a/src/db/slug-repository.ts
+++ b/src/db/slug-repository.ts
@@ -91,6 +91,17 @@ export class SlugRepository {
     const row = await SlugRepository.findByValue(db, slug);
     if (!row) return null;
 
+    // A link has exactly one system-generated (is_custom = 0) slug for its
+    // whole lifetime, so it can never stand in as "another auto slug" for
+    // itself: the primary-fallback query below promotes by `is_custom = 0`,
+    // and disabling the one and only such row would have that query
+    // re-select the very row being demoted next, leaving the link with no
+    // primary slug at all. The only current caller (disableSlug in
+    // link-management.ts) already refuses this before reaching here; this
+    // guard makes the repository enforce its own invariant rather than rely
+    // solely on that caller.
+    if (!row.is_custom) return null;
+
     // Disable and the primary fallback run in one transactional batch so a
     // failure cannot strand the link without a primary slug. The handover
     // re-checks inside the batch that this slug still holds primary: a
--- a/src/qr.ts
+++ b/src/qr.ts
@@ -55,11 +55,16 @@ export function makeQR(text: string): boolean[][] | null {
     ];
     const aligns = alignTable[ver];
     if (Array.isArray(aligns)) {
+      const last = aligns.length - 1;
       for (let ai = 0; ai < aligns.length; ai++) {
         for (let aj = 0; aj < aligns.length; aj++) {
+          // Skip the three positions that fall inside a finder pattern. Only
+          // those: from version 7 on, alignment patterns also sit on the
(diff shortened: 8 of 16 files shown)


DennisAlund:
**Flag for judgment: QR encoder uses wrong EC-codeword totals for versions 6-10, and never emits version-info blocks for 7-10** (`src/qr.ts`)

```ts
const eccL = [0, 7, 10, 15, 20, 26, 18, 20, 24, 30, 18];
const totalCodewords = [0, 26, 44, 70, 100, 134, 172, 196, 242, 292, 346];
const numEcc = eccL[ver];
const numData = totalCodewords[ver] - numEcc;
```

`eccL` is the standard "EC codewords **per block**, level L" table. The code uses it directly as the **total** EC codeword count without multiplying by the number of blocks (1 block for versions 1-5, but 2 for versions 6-9 and 4 for version 10). For versions 1-5 this happens to be correct (1 block), but for versions 6-10 the real EC totals should be 36, 40, 48, 60, 72; the code uses 18, 20, 24, 30, 18 instead. The encoder also never writes the two mandatory 6×3 version-information blocks the spec requires for versions 7-10.

Reachable: custom slugs are allowed up to 128 characters, so `origin + slug + "?utm_medium=qr"` easily exceeds the 106-byte version-5 boundary for a long custom slug, and the QR endpoint (`src/api/qr.ts`, `src/mcp/server.ts`) returns 200 with a normal-looking SVG that most real-world scanners will likely fail to decode.

Not fixing directly: a correct fix means implementing block splitting/interleaving per the spec plus version-info emission, which is a non-trivial rewrite of the hand-rolled encoder rather than a small, low-risk patch. Needs a decision on whether to fix the encoder properly, cap supported slug/URL length to stay within versions 1-5, or accept the current limitation.

---

**Flag for judgment: unrestricted server-side fetch of user-supplied link URLs for auto-labeling (SSRF pattern)** (`src/title-fetch.ts`, `src/services/link-management.ts`, `src/api/links.ts`)

`createLink` only validates `parsed.protocol === "http:" || "https:"`; it does not restrict host or IP. Any caller holding only the `create` API scope can create a link to `http://127.0.0.1/...` or another internal/link-local address, and the Worker fetches it server-side (following redirects) via `fetchPageTitle`, storing the resulting `<title>` as the link's label — visible via `GET /links/:id`.

This is a standard SSRF shape: no allow/deny policy exists in the codebase for the destination host. Cloudflare Workers' own network isolation (no direct route to RFC1918/loopback addresses unless the deployment adds a Tunnel or service binding) limits practical reach in most deployments, but nothing in this code enforces that — it depends entirely on platform topology.

Not fixing directly: needs a product decision on policy (block RFC1918/loopback/link-local resolution, whether to also block redirects into such ranges, whether this should apply to all outbound fetches or just this one) rather than a one-line fix.

---

DennisAlund:
Both items above are now addressed on this branch, each with its own regression test and CI green on the resulting head:

- QR encoder: fixed in 9bde31a (per-version block splitting/interleaving, version-info blocks for v7-10, alignment-pattern reservation), verified against a real decoder (jsQR).
- title-fetch SSRF: fixed in bf4d709c (`isPublicHttpUrl()` gate plus per-redirect-hop vetting).

Leaving this comment as the record of what the original findings were and how they were resolved.

---

src/title-fetch.ts
```diff
@@ -1,17 +1,154 @@
...
+    return groups !== null && !isReservedIPv6(groups);
+  }
+
+  // URL parsing already normalized decimal, octal and hex IPv4 spellings
+  // (2130706433, 0x7f.0.0.1, 0177.0.0.1) into dotted quads, so one check
+  // on the normalized form covers every spelling.
+  const octets = parseIPv4(host);
+  if (octets) return !isReservedIPv4(octets);
+
+  if (host === "localhost" || host.endsWith(".localhost")) return false;
```
DennisAlund, line 45: **A trailing dot bypasses every name-based block in this guard.**

The WHATWG URL parser keeps a fully qualified hostname verbatim, so `new URL("http://metadata.google.internal./computeMetadata/v1/").hostname` is `"metadata.google.internal."`. That misses `host === "localhost"`, `.endsWith(".localhost")`, `.endsWith(".internal")` and `.endsWith(".local")` alike, while DNS resolves `name.` to the same address as `name`.

Running the shipped function, all four of these return `true`:

```
http://metadata.google.internal./computeMetadata/v1/
http://localhost./
http://db.localhost./
http://printer.local./
```

The cloud metadata endpoint the doc comment names is reachable through the guard.

Fix: strip one trailing `.` from `host` before the comparisons. Numeric literals need no change, the parser already normalizes `http://127.0.0.1./` to `127.0.0.1`.

DennisAlund (commented):
## Code review: PR #59

Reviewed all 6 commits at `bf4d709`. 14 findings, 12 filed inline below. The two that follow could not be anchored: one file is untouched by this PR, the other line falls outside the diff hunks.

I verified the headline findings by running code rather than reading it: decoded QR payloads of every length 1 to 271 with jsQR, ran `isPublicHttpUrl` against 16 probe URLs, and checked `Uri.parse` against the real Dart toolchain.

### Highest priority

1. **SSRF guard, trailing-dot bypass** (`src/title-fetch.ts:45`). `http://metadata.google.internal./` passes.
2. **SSRF guard, IPv6 gaps** (`src/title-fetch.ts:105`). `::a.b.c.d` and `2002::/16` wrap loopback and link-local addresses past the check.
3. **Dart SDK parity** (below). The defect this PR fixed in Python is still live in Dart.
4. **The custom-slug toast is destroyed by the redirect that follows it** (`src/client.ts:280`). The PR's headline fix does not reach the user.

### Not anchorable, 1: Dart SDK carries the same `InvalidURL` escape, so the single-SDK rationale does not hold

`sdk/dart/lib/src/base_client.dart:170`. CLAUDE.md: "Any change to one SDK under `sdk/` must be evaluated and applied to the others." The commit message argues there is nothing to port. There is.

`_buildUri()` calls `Uri.parse('$_baseUrl$path')` at lines 172, 177 and 181, outside any `try`. Both call sites (`requestJson` line 60, `requestText` line 131) invoke it *before* the `try { streamed = await _httpClient.send(request); } catch (e) { throw ShrtnrError(0, ...) }` block that starts at line 77.

Confirmed against dart 3.10.4, using the exact input the new Python test uses:

```
Uri.parse('https://example.com:notaport/links')
  -> FormatException: Invalid port (at character 21)
```

So `ShrtnrClient(baseUrl: 'https://example.com:notaport', apiKey: k).links.list()` throws a raw `FormatException`, breaking `sdk/dart/README.md:142`: "Network failures also throw `ShrtnrError` with `status: 0`."

Fix: wrap the `_buildUri` call in both `requestJson` and `requestText`. The TypeScript SDK is worth the same check before this merges.

### Not anchorable, 2: non-ASCII QR payloads

Filed as a file-level comment on `src/qr.ts`, because the offending line 13 sits outside this PR's hunks.

### Process gate

CLAUDE.md: "Any change under `src/pages/`, `src/admin/`, `src/client.ts`, `src/styles.ts` or `src/index.tsx` admin routes ends with `yarn e2e` green, not only `yarn test`. A code review or release that touches those paths is not complete until the e2e suite has run."

This PR changes `src/client.ts` and reports 15 e2e failures. Reproducing on `main` is not the same as green. The gap is load-bearing here: finding 4 above is exactly the class of defect the rule exists for, since a render test cannot see a toast that a navigation destroys. `e2e/links.spec.ts` also has no `m-custom` step at all, so the changed create path has no browser coverage.

### Remaining findings, by weight

- `src/title-fetch.ts:140`: redirect loop gives each hop its own 5s timeout, raising the worst case to 30s of attacker-controlled `waitUntil`.
- `src/db/slug-repository.ts:103`: the guard blocks `disable()` instead of fixing the fallback query, and overloads the `null` return.
- `src/qr.ts:236`: GF(256) tables rebuilt per Reed-Solomon block, now 2 to 4 times per code.
- `sdk/python/src/shrtnr/resources/links.py:55`: the same `_request` body copy-pasted into 8 methods, with `_base.py` sitting right there.
- `src/__tests__/unit/qr.test.ts:143`: the version-information test recomputes the writer's own formula.
- `src/__tests__/unit/client-custom-slug-toast.test.ts:16`: `extractTopLevelChunk` now duplicated four times.
- `sdk/python/tests/test_async_client.py:95`: the new async test never closes its client.
- `src/__tests__/unit/client-t-replacer.test.ts:11`: em dash in a new comment.

src/title-fetch.ts
```diff
@@ -1,17 +1,154 @@
...
+  const rest = halves.length === 2 ? toGroups(halves[1]) : [];
+  if (!head || !rest) return null;
+  if (halves.length === 1) return head.length === 8 ? head : null;
+  const missing = 8 - head.length - rest.length;
+  if (missing < 1) return null;
+  return [...head, ...new Array(missing).fill(0), ...rest];
+}
+
+function isReservedIPv6(g: number[]): boolean {
+  const leadingZero = g.slice(0, 5).every((x) => x === 0);
```
DennisAlund, line 105: **`isReservedIPv6` misses IPv4-compatible (`::a.b.c.d`) and 6to4 (`2002::/16`) addresses.**

Both forms carry an IPv4 address that the guard never judges, so private and loopback targets pass when written that way. Verified against the shipped function:

| Input | Wraps | `isPublicHttpUrl` |
|---|---|---|
| `http://[::127.0.0.1]/` | 127.0.0.1 | `true` |
| `http://[::a00:1]/` | 10.0.0.1 | `true` |
| `http://[2002:7f00:1::]/` | 127.0.0.1 | `true` |
| `http://[2002:a9fe:a9fe::]/` | 169.254.169.254 | `true` |

The parser normalizes `[::127.0.0.1]` to `[::7f00:1]`, so the `leadingZero` branch on line 106 only matches `::` and `::1`, and `g[5]` is `0` rather than `0xffff`. Neither special case fires and the function falls through to `return false`.

The new 27-entry blocked-target list covers `::ffff:` and `64:ff9b::` but not `::/96` or `2002::/16`.

Fix: when `leadingZero && g[5] === 0 && (g[6] | g[7]) !== 0`, judge `groupsToIPv4(g[6], g[7])`; add a `g[0] === 0x2002` branch doing the same on `g[1]`/`g[2]`.

src/title-fetch.ts
```diff
@@ -1,17 +1,154 @@
...
-      signal: AbortSignal.timeout(5000),
-    });
+    let current = url;
+    let res: Response | null = null;
+    for (let hop = 0; hop <= MAX_REDIRECTS; hop++) {
+      if (!isPublicHttpUrl(current)) return null;
+      const candidate = await fetch(current, {
+        headers: { "User-Agent": "Shrtnr/1.0 (link preview)" },
+        redirect: "manual",
+        signal: AbortSignal.timeout(5000),
```
DennisAlund, line 140: **The per-hop timeout raises the worst case from 5s to 30s.**

The previous single `fetch(url, { redirect: "follow", signal: AbortSignal.timeout(5000) })` bounded the entire followed chain at 5 seconds. This loop builds a fresh signal on each of the 6 iterations (`hop = 0..MAX_REDIRECTS`), so each hop gets its own budget.

A server that answers every request with a 302 to a new public URL after 4.9s holds `autoLabelLink`'s `waitUntil` for roughly 30 seconds. Every link creation goes through this path (`src/api/links.ts:85` and `:507`), and the destination is user-supplied.

Fix: hoist one `AbortSignal.timeout(5000)`, or a deadline computed once, outside the loop and pass it to every hop.

src/client.ts
```diff
@@ -272,6 +267,24 @@ function createLink() {
...
+// slug itself was accepted (it may already be taken) separately from the
+// link creation that already succeeded by the time this runs.
+function attachCustomSlugAndGo(linkId, slug) {
+  api('/links/' + linkId + '/slugs', { method: 'POST', body: JSON.stringify({ slug: slug }) }).then(function(slugRes) {
+    closeModal();
+    if (!slugRes.ok) {
+      return slugRes.json().then(function(data) {
+        toast(data.error || t('client.customError'), 'error');
+      }).catch(function() { toast(t('client.customError'), 'error'); }).then(function() {
+        window.location.href = '/_/admin/links/' + linkId;
```
DennisAlund, line 280: **The error branch navigates away, so the toast this PR adds never gets read.**

`toast()` writes into the in-page `#toast` element with a 3000ms lifetime. Assigning `window.location.href` on the next microtask replaces the document and takes the toast with it. The user still lands on the detail page carrying only the auto-generated slug, with no durable sign that the slug they typed was rejected. That is the failure this PR set out to surface.

The three new vitest cases pass because they stub `window` as a plain object, so the navigation is never modelled.

`doAddSlug`, the handler this function was modelled on, deliberately does not navigate or close the modal on failure. Mirror that: on the error branch skip both `closeModal()` (line 275) and this redirect, and leave the user in the modal with the message visible.

src/qr.ts
```diff
@@ -161,6 +187,51 @@ function toBin(n: number, len: number): string {
...
+    }
+  }
+  for (let i = 0; i < eccPerBlock; i++) {
+    for (const block of blocks) out.push(block.ecc[i]);
+  }
+  return out;
+}
+
 function rsEncode(data: number[], numEcc: number): number[] {
   const exp = new Uint8Array(512);
```
DennisAlund, line 236: **`rsEncode` rebuilds the GF(256) tables on every call, and block splitting now multiplies that.**

Each call allocates `new Uint8Array(512)` plus `new Uint8Array(256)`, runs 255 + 257 iterations to fill them, then builds the generator polynomial in O(numEcc²). `interleaveBlocks` now calls this once per block: 2 calls for versions 6 to 9 and 4 for version 10, where the old code made exactly 1. `src/api/qr.ts` serves this per request.

Fix: hoist `exp`, `log` and `gfMul` to module scope, and memoize the generator polynomial by `numEcc`.

src/db/slug-repository.ts
```diff
@@ -91,6 +91,17 @@ export class SlugRepository {
...
+    // A link has exactly one system-generated (is_custom = 0) slug for its
+    // whole lifetime, so it can never stand in as "another auto slug" for
+    // itself: the primary-fallback query below promotes by `is_custom = 0`,
+    // and disabling the one and only such row would have that query
+    // re-select the very row being demoted next, leaving the link with no
+    // primary slug at all. The only current caller (disableSlug in
+    // link-management.ts) already refuses this before reaching here; this
+    // guard makes the repository enforce its own invariant rather than rely
+    // solely on that caller.
+    if (!row.is_custom) return null;
```
DennisAlund, line 103: **This guard forbids the operation instead of fixing the query that makes it unsafe, and it overloads the `null` return.**

The comment above states the root cause precisely: the primary-fallback `UPDATE slugs SET is_primary = 1 WHERE link_id = ? AND is_custom = 0` can re-select the very row being demoted two statements later. Adding `AND slug != ?` to that query fixes the invariant for every caller, present and future, and leaves `disable()` usable on auto slugs.

The guard chosen here instead makes the repository refuse a call whose only defect lives in the query, and it returns `null`, the same value a `findByValue` miss returns. `disableSlug` (`src/services/link-management.ts:432`) maps `null` to `fail(404, "Slug not found")`, so a future direct caller gets "Slug not found" for a slug that plainly exists.

`remove()` (line 165) carries the identical un-excluded fallback query and would need the same treatment either way.

sdk/python/src/shrtnr/resources/links.py
```diff
@@ -52,16 +52,22 @@ def _url(self, path: str, query: dict[str, str | None] | None = None) -> str:
         return _build_url(self._base_url, path, query)
 
     def _request(self, method: str, url: str, **kwargs: Any) -> Any:
+        # httpx.InvalidURL isn't a RequestError subclass; catch it too so a
```
DennisAlund, line 55: **This fix is copy-pasted into 8 method bodies when `_base.py` already exists for exactly this.**

`LinksResource._request`, `LinksResource._request_text`, `AsyncLinksResource._request`, `AsyncLinksResource._request_text` and the four Bundles/Slugs equivalents are now byte-identical bodies carrying byte-identical three-line comments. `_base.py` already holds `_build_url`, `parse_json_response` and `parse_text_response`, the shared pieces of this same request path.

The next httpx exception-hierarchy surprise, or a retry policy, a header, a timeout, has to land in 8 places again and can silently miss one.

Fix: move `_request` and `_request_text` into a `_BaseResource` / `_AsyncBaseResource` in `_base.py` and have the three resource classes inherit.

src/__tests__/unit/qr.test.ts
```diff
@@ -72,3 +72,94 @@ describe("makeQR", () => {
...
+describe("makeQR version information", () => {
+  // 18-bit version information (6 data bits + 12 BCH bits), ISO/IEC 18004 Table D.1.
+  const VERSION_INFO: Record<number, number> = {
+    7: 0x07c94,
+    8: 0x085bc,
+    9: 0x09a99,
+    10: 0x0a4d3,
+  };
+
+  function readVersionInfo(grid: boolean[][]): { topRight: number; bottomLeft: number } {
```
DennisAlund, line 143: **This test asserts the writer against itself, so a placement error would still pass.**

`readVersionInfo` recomputes `a = size - 11 + (i % 3)` and `b = Math.floor(i / 3)`, byte for byte the formula `makeQR` uses to write the version blocks, then reads `grid[b][a]` and `grid[a][b]` exactly as the writer wrote them. Transpose the two blocks or shift them by a module and this assertion still passes. The new v7+ blocks are covered for value, not for position.

The scannability suite cannot close the gap either: jsQR derives the version from the symbol dimension and only falls back to the version bits, so it decodes fine even with wrong version information.

Fix: assert against literal module coordinates from ISO/IEC 18004 Figure 25, or compare against a known-good reference symbol.

src/__tests__/unit/client-custom-slug-toast.test.ts
```diff
@@ -0,0 +1,136 @@
...
+// "Link created" toast either way, so a rejected custom slug (already
+// taken, invalid) failed silently: the user was told the slug they asked
+// for was attached when it wasn't. This drives createLink() against a
+// stubbed api() that succeeds on the first call and fails on the second,
+// and asserts the failure actually surfaces.
+import { describe, expect, it } from "vitest";
+import { adminClientScript } from "../../client";
+import type { Translations } from "../../i18n/types";
+
+function extractTopLevelChunk(source: string, startPattern: RegExp): string {
```
DennisAlund, line 16: **`extractTopLevelChunk` is now copied into four test files.**

The same 12-line function, down to the `/^(function |var |if |window\.|document\.)/` terminator heuristic, already lives at `src/__tests__/unit/client-api-error-toasts.test.ts:59` and `src/__tests__/unit/client-script-escaping.test.ts:13`. This PR adds copies here and in `client-t-replacer.test.ts:16`.

The heuristic is fragile: it breaks the moment a helper in `client.ts` starts with an unindented `const` or `let`, and the repair would then have to land in four places.

Fix: extract it to a shared helper module under `src/__tests__/`.

sdk/python/tests/test_async_client.py
```diff
@@ -92,6 +92,17 @@ async def test_async_network_error_wraps_as_status_0(client: AsyncShrtnr) -> Non
         assert exc_info.value.status == 0
 
 
+async def test_async_malformed_base_url_wraps_as_status_0() -> None:
```
DennisAlund, line 95: **This test never closes its client, breaking the teardown discipline the rest of the file follows.**

The module's `client` fixture (line 33) wraps its yield in `try/finally: await c.aclose()` precisely so the underlying `httpx.AsyncClient` connection pool is released. This test builds `AsyncShrtnr(...)` directly on line 100 and returns after the `pytest.raises` block, leaving the pool open for the remainder of the session. That risks an "unclosed client" `ResourceWarning` or an event-loop-closed error in CI.

Fix: wrap it in `async with`, or add `await client.aclose()`.

The sync twin in `test_client.py` is worth the same check.

src/__tests__/unit/client-t-replacer.test.ts
```diff
@@ -0,0 +1,60 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+// The admin client script's t() interpolates {param} placeholders into a
+// translation string. src/i18n/index.ts's createTranslateFn() already
+// guards this with a replacer function instead of a plain string, because
+// String.prototype.replace treats "$&", "$`", "$'", "$1" etc. in a string
+// replacement as special patterns instead of literal text. t() in
+// client.ts reimplements the same interpolation client-side and needs the
+// same guard, or a param value containing one of those sequences (an API
+// key title, a bundle name — both free text) corrupts the rendered string.
```
DennisAlund, line 11: **Em dash in new comment text.**

CLAUDE.md, Writing rules, applies to all produced material including comments: "No em dashes. Use colon, comma, or period."

This line reads `// key title, a bundle name — both free text) corrupts the rendered string.` A comma or a colon works here.

src/qr.ts
DennisAlund, line 1: **`makeQR` treats input as Latin-1 and sizes the symbol by `text.length`, so non-ASCII payloads produce silently undecodable codes.**

Line 13 runs `data.push(text.charCodeAt(i))`, which yields `233` for `é` (Latin-1, where UTF-8 needs `0xC3 0xA9`) and `20320` for `你`. For any code point above 255 the value then overflows the byte field: `toBin(20320, 8)` returns a 15-character string and desynchronizes the whole bit stream. The capacity check on line 17 compares against `text.length` rather than the encoded byte length, so the version is undersized too.

Decoding real output with jsQR:

- `makeQR("https://s.ex/café")` returns a version-1 grid that decodes to `""`.
- `makeQR("https://s.ex/你好")` returns a grid jsQR cannot read at all.

Neither returns `null`, so callers ship a broken image. This is reachable today through the MCP `qr` tool (`src/mcp/server.ts:491`), whose `base_url` is caller-supplied and validated only by `z.string().url()`, which accepts IDN hosts.

Fix: encode with `TextEncoder` and derive the version from the byte length.

(Filed against the file rather than a line, because line 13 falls outside this PR's hunks.)

src/title-fetch.ts
DennisAlund: Fixed in d747cab. `hostname.toLowerCase().replace(/\.$/, "")` runs before every comparison, so `metadata.google.internal.`, `localhost.`, `db.localhost.` and `printer.local.` now hit the same branches as the bare names. All four are in the blocked-target list and failed before the fix. An allow-side case pins `https://example.com./` as still public so the strip cannot over-block.

src/title-fetch.ts
DennisAlund: Fixed in d747cab, both branches as you laid them out. `leadingZero && g[5] === 0 && (g[6] | g[7]) !== 0` judges `groupsToIPv4(g[6], g[7])`, with the `!== 0` keeping `::` and `::1` on the existing branch; `g[0] === 0x2002` judges `groupsToIPv4(g[1], g[2])`. All four probe inputs from the table are in the blocked list and failed before the fix. `http://[2002:5db8:d822::1]/` (wrapping 93.184.216.34) is pinned as public so the 6to4 branch judges the address rather than the prefix.

src/title-fetch.ts
DennisAlund: Fixed in 82c46b7. One `AbortSignal.timeout(5000)` is built before the loop and passed to every hop, so the whole chain shares the 5s budget. The test drives a two-redirect chain and asserts all three `fetch` calls received the same signal instance; it failed before the hoist.

src/client.ts
DennisAlund: Correct about the navigation destroying the toast. caf23c8 made the failure branch skip both `closeModal()` and the redirect, mirroring `doAddSlug`, with tests asserting `location.href` untouched and the modal still open.

Then 28d603c removed the whole path. `showCreateModal()` has no caller: nothing in `src/pages/` renders an onclick for it, no listener wires it, and a Playwright pass over every admin page found the function defined and zero elements referencing it. The only reachable create path is `quickShorten`, which never sends a custom slug, so `createLink` and `attachCustomSlugAndGo` were dead and this defect could not reach a user. Deletion was the call made in review; the vitest file, the `createLink` row in the error-toast table, and nine orphaned translation keys went with it.

src/qr.ts
DennisAlund: Fixed in 0fda7e5. `GF_EXP`, `GF_LOG` and `gfMul` sit at module scope and build once per isolate. The generator polynomial is memoized by ECC count in a module-level `Map`; every block in a symbol carries the same count, so the first block of a code builds it and the rest reuse it. The cache holds pure, deterministic data keyed by a small integer, not request state, so it is on the right side of the module-level-state rule for Workers.

src/db/slug-repository.ts
DennisAlund: Split decision, in a40e516.

The `null` overload is real and fixed: the guard now throws `Error("Cannot disable the system-generated slug ...")`, so a direct caller can no longer be told "Slug not found" about a row that exists. The test asserts the rejection and that the primary flag is untouched.

I kept the guard rather than adding `AND slug != ?` to the fallback query, because that clause does not rescue the call. A link has exactly one `is_custom = 0` slug. Excluding the row being demoted leaves the promote matching nothing, and the demote that follows still sets `is_primary = 0` on the only primary. The link lands on zero primaries either way; the clause changes the path, not the outcome. Refusing the call is the only thing that holds the invariant. The comment above the guard now records this reasoning so the next reader does not reopen it. `remove()` already carries the same guard, so it needs no change.

sdk/python/src/shrtnr/resources/links.py
DennisAlund: Fixed in 1f5b39d. `_SyncResource` and `_AsyncResource` in `shrtnr/_base.py` (the module lives one level up from `resources/`) now hold construction, `_headers`, `_json_headers`, `_url` and the request path. `_send` is the single place that maps `(httpx.RequestError, httpx.InvalidURL)` to `ShrtnrError(0, ...)`; `_request` and `_request_text` wrap it. Links, Bundles and Slugs inherit and keep only endpoint methods. Net 107 lines out of the resource modules. 104 tests, mypy and ruff clean.

src/__tests__/unit/qr.test.ts
DennisAlund: Fixed in 453b488. The reader is gone. Each version now has a literal 6x3 module picture and the literal columns it occupies (v7: cols 34-36, v8: 38-40, v9: 42-44, v10: 46-48), laid out per Figure 25 from the Table D.1 values; the bottom-left test reads the transposed coordinates against the same picture. Verified by mutation: shifting the block one column, and transposing the bit order, each fail the new assertions and passed the old one.

src/__tests__/unit/client-custom-slug-toast.test.ts
DennisAlund: Fixed in cdb179e. `extractTopLevelChunk` lives in `src/__tests__/client-script.ts` and the four files import it. The terminator list is widened to `const`, `let` and `class` so the failure mode you describe cannot arrive silently, and the boundary rule is documented at the definition. This particular test file was later removed in 28d603c along with the dead modal it covered; the other three importers remain.

sdk/python/tests/test_async_client.py
DennisAlund: Fixed in 1f5b39d. Both malformed-base-url tests now build their client with the context manager (`async with AsyncShrtnr(...)`, `with Shrtnr(...)`). On the sync twin: its module fixture returns without closing, so that file had no teardown habit to break, but a client built inside a test should still release its pool, so it got the same treatment. Suite passes with `-W error::ResourceWarning`.

src/__tests__/unit/client-t-replacer.test.ts
DennisAlund: Fixed in cdb179e: comma. Scanned the rest of the branch diff for em dashes; the only remaining one is in `sdk/python/src/shrtnr/_base.py:96`, which predates this branch on `main`, so I left it for a separate change.

src/qr.ts
DennisAlund: Reproduced and fixed in c177740. With the suite's own rasterizer: `https://s.ex/café` gave a version-1 grid decoding to `""`, `https://s.ex/你好` was undecodable, neither returned null. `makeQR` now encodes with `TextEncoder`, and the existing capacity check reads the encoded length, so version choice and the 271-byte ceiling are byte-based. Tests round-trip accented, CJK and emoji payloads through jsQR, pin `"好".repeat(60)` (180 bytes) to version 8 rather than the length-based version 3, and check that 91 three-byte characters (273 bytes) return null while 90 (270 bytes) encode.

DennisAlund:
## Review follow-up: all 14 findings addressed, 11 commits at `28d603c`

Every inline finding has a reply in its thread with the fixing commit. The two that could not be anchored, and the process gate:

### Dart SDK parity (not anchorable, 1)

Confirmed and fixed in e8bea78. `Uri.parse('https://example.com:notaport/links')` throws `FormatException: Invalid port (at character 21)` on dart 3.10.4, and `_buildUri` ran outside the `try` in both `requestJson` and `requestText`. The wrap now lives inside `_buildUri`, so both paths and any later one are covered from one place. The test exercises `links.list()` (requestJson) and `links.qr()` (requestText).

The TypeScript SDK was checked as asked and needs no change. `buildUrl` concatenates strings without parsing, so the malformed URL reaches `fetch` inside its `try` and already surfaces as `ShrtnrError(0, "Failed to parse URL ...")`. Verified by running it, not by reading it.

### Non-ASCII QR payloads (not anchorable, 2)

Reproduced and fixed in c177740; details in the file-level thread.

### Process gate

`yarn e2e` is 17/17 green on this machine, on every run. The 15 failures the PR body reported were an artifact of that sandbox's network policy and do not exist here.

The `m-custom` e2e step could not be written, for a reason that changed the shape of this PR: **`showCreateModal()` has no caller.** No onclick in `src/pages/`, no listener, no shortcut. A Playwright pass over every admin page found the function defined and zero elements referencing it. The only reachable create path is `quickShorten`, which never sends a custom slug. So the PR's headline fix and inline finding 4 both sat on dead code, and no browser test can click a control the app never renders. Per the decision in review, 28d603c removes `showCreateModal`, `createLink`, `attachCustomSlugAndGo`, their vitest file, the `createLink` row in the shared error-toast table, and nine orphaned translation keys across en, sv and id. The PR now carries no live change to `client.ts` beyond that deletion.

### One pushback

`slug-repository.ts:103`: the `null` overload is fixed (it throws now), but the guard stays. `AND slug != ?` does not hold the invariant: with exactly one `is_custom = 0` slug per link, excluding the demoted row leaves the promote matching nothing and the demote still lands the link on zero primaries. Reasoning is in the thread and in the code comment.

### Verification

- `yarn test`: 1375 green. `yarn e2e`: 17 green. `tsc --noEmit` clean.
- Python: 104 green, mypy and ruff clean. Dart: 97 green with `--exclude-tags e2e` as CI runs it, `analyze --fatal-infos` clean. TypeScript: 84 green.
- All three `x-spec-hash` values match the live spec; no drift to record.
- Workers lens: `redirect: "manual"` returns the 3xx "as-is" per the Request docs, which is what the hand-followed loop relies on; the chain is bounded at 5s, inside `waitUntil`'s limit; `qr.ts`'s module-level tables are deterministic memoization, not request state.

### Worth knowing, not changed here

- One unreproduced flake in `src/__tests__/handler/api-links.test.ts` (Custom Slug Redirect area) on an early full-suite run: not seen again in 8 isolated and 10 full-suite runs, `main` clean over 5 runs, and this branch touches neither that file nor `src/api/`.
- Pre-existing: asking for a custom slug on a URL that already exists silently drops the slug (`quickShorten`'s duplicate branch redirects without attaching it). Same class as the bug this PR started from, outside its scope.

DennisAlund:
`Worker (yarn test)` failed twice in a row on this head (run 35606699315, attempts 1 and 2), both times on the identical assertion: `src/__tests__/handler/api-links.test.ts > Custom Slug Redirect > should 301 redirect via a custom slug`, `Error: Test timed out in 5000ms` at line 700.

Not this PR's: `api-links.test.ts` predates this branch (present since `6b136c1`) and this branch touches neither that file nor `src/api/links.ts`.

Root cause, as far as I can tell without modifying the test: the test creates a link to a real, unmocked `https://custom-target.com` with no `label`, which triggers `autoLabelLink` → `fetchPageTitle`. `@cloudflare/vitest-pool-workers` drains a request's `ctx.waitUntil()` promises before the test's `SELF.fetch()` call resolves, so the test's first `await` sits on that fetch completing. `fetchPageTitle`'s own network operation is bounded by `AbortSignal.timeout(5000)` — the same 5000ms as vitest's default `testTimeout` — so if `custom-target.com` doesn't resolve/respond quickly on the runner's network, the two timers race almost exactly and either can fire first. This 5s cap predates this PR (the pre-review code had the same single `AbortSignal.timeout(5000)`); this branch's redirect-hop changes don't shorten or lengthen the overall bound. That matches what's reported above: clean over 8 isolated + 10 full-suite runs plus 5 clean `main` runs on a different machine, now failing twice in a row specifically on this GitHub Actions runner's network path to that domain.

I don't have permission to re-run the failed job here (403 on `rerun_failed_jobs`), and CLAUDE.md's testing rule ("Never modify or remove tests to accommodate code changes... only adding new tests is permitted") means I won't touch `api-links.test.ts` myself to mock that fetch. Flagging rather than guessing: the durable fix is mocking `fetchPageTitle`'s outbound call in that test (or giving the assertion a longer/more deterministic timeout), in a follow-up unrelated to this PR's diff.

---

DennisAlund:
## CI failure: root cause and fix (e6dd0fb)

The failed job was `Worker (yarn test)`, one test: `api-links.test.ts › Custom Slug Redirect › should 301 redirect via a custom slug`, timed out at 5s after 11.6s. This is the flake I flagged in the summary above; CI reproduced it and its log shows the mechanism.

**Cause.** The suite was making real outbound HTTPS requests from inside workerd. Every link a handler test creates schedules `autoLabelLink()` in `waitUntil`, which calls `fetchPageTitle()` on the link's URL, and nothing stubbed `fetch`. The job log has `DNS lookup failed ... cached.example.com`, `TLS peer's certificate is not trusted`, an `OPENSSL_internal:TLSV1_ALERT_INTERNAL_ERROR`, and one handshake to `custom-target.com` (a real parked domain) that stalled past the test timeout. Because `SELF.fetch` does not wait on `waitUntil`, the slow fetch also outlived its test and wrote a label into the next test's rolled-back storage, which is the uncaught `D1_ERROR: FOREIGN KEY constraint failed` in the same log.

This predates the branch (`main` fetched with `redirect: "follow"` and the same 5s signal). It stayed hidden because a healthy network answers those hosts in milliseconds: my 14 local runs contained none of that workerd noise, which is why 10+ green runs here were not evidence of anything. The CI runner's slower path to parked domains is what exposed it.

**Fix (e6dd0fb).** `vitest.config.mts` now sets miniflare's `outboundService` to a handler that answers every outbound fetch immediately with a `text/plain` body, so `fetchPageTitle()` returns `null` and `autoLabelLink()` writes nothing. Verified against the installed miniflare README (documents the plain handler form) and the pool's `SourcelessWorkerOptions`, which passes the field through. No prod code and no existing test changed.

**Guard.** `src/__tests__/handler/outbound-hermetic.test.ts` pins the wiring: a global `fetch()` to a real host must return the stub body, and `fetchPageTitle()` must return `null` in under a second. With the config stashed, the first case fails after 7.8s of real network time, the same shape as the CI failure.

**Verification.** `yarn test` 1377 green on three consecutive runs, zero DNS/TLS/FK lines in the workerd output. `tsc --noEmit` clean.

PR #60, opened 2026-09-18
Weekly review batch: URL-normalization corruption, silent theme-save failure, cross-SDK JSON-scalar-body crash

DennisAlund:
Weekly defect-hunting review of the app, its API, and all three SDKs. Six parallel passes (API/DB/services, admin UI, MCP/auth, TypeScript SDK, Python SDK, Dart SDK), each finding verified against the actual code and a failing-before/passing-after regression test before being fixed.

Note: PR #59 (last week's review batch) is still open and unmerged, so this branch was cut from the same `main` it targets and does not overlap with its fixes; the two are independent.

## Fixes

**`normalizeUrl()` stripped a trailing `?`/`#` that was part of the URL's own content** (`src/normalize-url.ts`)
- Wrong: it decided a trailing `?`/`#` was a redundant empty delimiter by checking only whether the whole string *ended* with that character, not whether it was the *first* occurrence acting as the actual delimiter.
- Impact: `normalizeUrl` runs on every link URL before it's stored and used for real 301 redirects. `https://a.com/?x=y?` (query value `y?`) lost its trailing `?`, and `https://a.com/redirect?next=https://example.com/?` corrupted an embedded redirect target the same way. `https://a.com/#section#` lost a trailing `#` that was fragment content.
- Fix: check that the first `#`/`?` in the string is also the last character, not just that the string ends with it.
- Test: four new cases (query-value `?`, embedded-redirect `?`, fragment-content `#`), confirmed to fail on the old code and pass on the new.

**`setTheme()` toasted "Theme updated" even when the settings save failed** (`src/client.ts`)
- Wrong: unlike every sibling settings setter (`setDefaultRange`, `saveSettings`, `setFilterBots`, `setFilterSelfReferrers`), which all check `res.ok` before toasting, `setTheme` fired the PUT and toasted success unconditionally.
- Impact: a failed save (transient 5xx, expired session) looked identical to a successful one. The theme still applied in that one browser via a client-side cookie, but the choice was silently never persisted to the account's settings row.
- Fix: check `res.ok`, matching the sibling setters; reuses the existing `client.themeUpdated`/`client.settingsError` i18n keys.
- Test: extracts the real handler from the generated script and drives it against both a success and a failure response.

**All three SDKs crashed instead of raising `ShrtnrError` on a bare JSON scalar in a 2xx body** (`sdk/python/src/shrtnr/_base.py`, `sdk/typescript/src/internal/http.ts`, `sdk/dart/lib/src/base_client.dart`)
- Wrong: Python already guarded against a literal JSON `null` on a 2xx body (added in an earlier release), but the guard only checked for `None`, not other bare scalars (a number, string, or bool). TypeScript and Dart had no guard for either case.
- Impact: a broken proxy/CDN serving a stale or truncated 2xx body as a bare scalar caused, depending on the SDK and endpoint: a raw `AttributeError`/`TypeError`/null-check crash instead of the documented `ShrtnrError` (single-object endpoints), or a falsy scalar (`0`) silently swallowed into an empty list via `data or []` (Python's `list()` endpoints) instead of raising at all.
- Fix: widened the check in all three SDKs to reject any parsed body that isn't an object/dict or an array/list, matching what every resource method actually expects.
- Test: one regression test per SDK per failure mode (null and non-null scalar; falsy-scalar-swallowed-by-list() for Python), each confirmed to fail on the old code and pass on the new.

## Flagged for judgment (not fixed)

**Slug values spliced unescaped into inline `onclick` JS-string literals** (`src/client.ts:445,489,503,517` — `doSetPrimary`, `doDeleteSlug`, `doDisableSlug`, `doEnableSlug`)
- These four call sites concatenate a slug raw into a single-quoted JS string inside a dynamically-built `onclick` attribute, with no escaping — the same bug class the codebase otherwise deliberately hardens against elsewhere via a `data-*` attribute + `dataset` pattern (see the comments around `client.ts:463-466`, `:1288-1290`, `:1636-1637`, `:1649-1651`, `:1660-1662`).
- **Not currently exploitable**: `validateCustomSlug`/`validateRandomSlug` (`src/slugs.ts`) restrict every slug write path to `[a-z0-9-]`, which cannot contain a quote.
- Not fixed directly because it isn't a live defect today; flagging because it's an inconsistency with the project's own documented mitigation for this exact bug class, so any future loosening of slug validation turns it into a real stored-XSS primitive. Migrating the four call sites to the `data-*` pattern is a judgment call on scope, not a one-line fix.

**Duplicated, outdated `hono` dependency via `@modelcontextprotocol/sdk`** (`yarn.lock`)
- `yarn.lock` resolves two copies of `hono`: our direct `^4.13.5` dependency, and a separate `^4.11.4` pulled in transitively by `@modelcontextprotocol/sdk`, resolved to `4.12.27` — vulnerable to several advisories already patched in our own 4.13.5 (`yarn audit` flags them).
- **Confirmed not reachable in the deployed Worker**: the only place `@modelcontextprotocol/sdk` imports `hono` is `dist/{esm,cjs}/examples/server/honoWebStandardStreamableHttp.js`, a bundled example file. Our code only imports `@modelcontextprotocol/sdk/server/mcp.js`; the example (and its `hono`/`express`/`cors` dependencies) is never pulled into the import graph Wrangler bundles.
- Not fixed directly: it's dead weight in the dependency tree rather than a live vulnerability, and yarn classic's failure to dedupe two semver-compatible ranges (`^4.11.4` and `^4.13.5` both resolve within 4.x, so 4.13.5 satisfies both) is a lockfile-hygiene issue, not a defect in this diff's scope. Worth a `yarn dedupe` or a `resolutions` override at the maintainer's convenience.
- The remaining `yarn audit` findings (vite, postcss, qs, fast-uri, ip-address, nanoid, sharp, body-parser) are all transitive dev-only tooling (wrangler/Playwright build chain) that never ships in the Worker; not flagging individually.

## Verification
- `yarn test`: 1305 tests, all green.
- `yarn e2e`: 14 failures reproduce identically (same test names) on a clean `main` checkout in a side-by-side worktree in this sandbox — TLS cert-trust and browser-executable-version mismatches from the sandbox's own proxy/browser-cache setup, unrelated to this diff.
- Python SDK: 104 tests, all green. `ruff check` / `ruff format --check` / `mypy` clean on changed files.
- TypeScript SDK: 86 tests, all green. `tsc --noEmit` clean.
- Dart SDK: 95 tests, all green (the one remaining failure needs e2e env vars, unrelated). `dart analyze` clean on changed files.

---

https://claude.ai/code/session_01Gvz1NwUWgA2STuMUFNVzqD

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gvz1NwUWgA2STuMUFNVzqD)_


src/__tests__/unit/client-settings-toasts.test.ts
```diff
@@ -0,0 +1,80 @@
...
+// settings setter (setDefaultRange, saveSettings, setFilterBots,
+// setFilterSelfReferrers), which all check res.ok before deciding which
+// toast to show. This extracts the real handler from the generated script
+// and drives it against both a success and a failure response to prove the
+// toast now reflects the actual outcome.
+import { describe, expect, it } from "vitest";
+import { adminClientScript } from "../../client";
+import type { Translations } from "../../i18n/types";
+
+function extractTopLevelChunk(source: string, startPattern: RegExp): string {
```
DennisAlund: This re-implements `extractTopLevelChunk`, which main now exports from `src/__tests__/client-script.ts` (commit cdb179e via #59, merged after this branch was cut). That commit exists to remove the private copies, so merging this adds a fourth one back. This copy also lacks the `const |let |class ` terminators the shared version gained, so a future top-level `const` after `setTheme` would get swallowed into the chunk.

Replace lines 15 to 27 with:

```ts
import { extractTopLevelChunk } from "../client-script";
```

DennisAlund (commented):
Reviewed the merge result of this branch onto current main (which now includes #59). `yarn test` passes with 1382 tests and `yarn e2e` passes with all 17 Playwright steps, so the e2e failures noted in the description were sandbox-only. The three fixes are correct. Four items below before merge: one duplicate helper that undoes a commit already on main, one sibling setter with the same defect this PR fixes, two writing-rule hits in new text, plus one follow-up note that does not block.

src/__tests__/unit/client-settings-toasts.test.ts
```diff
@@ -0,0 +1,80 @@
+// Copyright 2026 Oddbit (https://oddbit.id)
+// SPDX-License-Identifier: Apache-2.0
+
+// setTheme() used to toast "Theme updated" unconditionally, regardless of
+// whether the PUT to /settings actually succeeded — unlike every sibling
```
DennisAlund: Writing rules in CLAUDE.md apply to comments and test names: no em dashes, no hollow intensifiers. This line has both ("actually succeeded — unlike"), and the test name on line 67 repeats "actually". Suggest "whether the PUT to /settings succeeded. Every sibling settings setter ..." and "toasts success only when the save succeeds".

sdk/python/src/shrtnr/_base.py
```diff
@@ -89,16 +89,19 @@ def parse_json_response(response: httpx.Response) -> Any:
...
-    if parsed is None:
-        raise ShrtnrError(response.status_code, "Response body is JSON null")
+    # A body that is valid JSON but parses to a bare scalar (null, a number,
+    # a string, or a bool) rather than an object or array (4 bytes for
+    # `null`, so it passes the empty-body check above, and valid JSON, so it
+    # passes the parse above) used to reach here unchanged. Every
+    # single-object resource method's `SomeModel.from_dict(...)` expects a
+    # dict and crashed on it with a bare AttributeError, and a falsy scalar
+    # (`0`, `false`, `""`) made list() endpoints silently return `[]` instead
+    # of raising, via `data or []` — the same failure mode the empty-body
```
DennisAlund: Em dash carried over from the old comment into the rewrite. A colon or period reads the same: "via `data or []`: the same failure mode the empty-body check exists to prevent".

src/client.ts
```diff
@@ -128,8 +128,10 @@ function applyTheme(theme) {
 function setTheme(theme) {
   applyTheme(theme);
   document.cookie = 'theme=' + theme + ';path=/;max-age=31536000;SameSite=Lax';
-  api('/settings', { method: 'PUT', body: JSON.stringify({ theme: theme }) });
-  toast(t('client.themeUpdated'));
+  api('/settings', { method: 'PUT', body: JSON.stringify({ theme: theme }) }).then(function(res) {
```
DennisAlund: The fix is right and matches the sibling setters. One sibling is left with the identical defect, though: `setLanguage` a few lines below (line 141) fires the PUT and reloads unconditionally, so a failed save still looks like a success and the account row never records the new language. Since this PR's rationale is parity with the other setters, `setLanguage` belongs in the same change: reload on `res.ok`, otherwise `toast(t('client.settingsError'), 'error')`. A vitest case next to the two theme cases would cover it.

sdk/python/src/shrtnr/_base.py
```diff
@@ -89,16 +89,19 @@ def parse_json_response(response: httpx.Response) -> Any:
...
+    # check exists to prevent, just reached from a non-empty body. Arrays
+    # and objects are left alone: they are the only shapes a resource method
+    # ever legitimately expects.
+    if not isinstance(parsed, (dict, list)):
```
DennisAlund: Follow-up, not blocking this PR. The guard now admits any dict or list, but the transport cannot know which one the caller expects. A 2xx body of `[]` on `links.get()` still reaches `Link.from_dict([])` and raises a bare `AttributeError`, and `{}` on `links.list()` iterates dict keys. Same shape in the TypeScript and Dart transports. The fix at the right depth is an expected-shape parameter on the request path (object vs array), applied to all three SDKs together. Worth an issue so it does not get lost.

src/__tests__/unit/client-settings-toasts.test.ts
DennisAlund: Fixed in 04f6e51: now imports `extractTopLevelChunk` from `../client-script` instead of the private copy.

---

src/__tests__/unit/client-settings-toasts.test.ts
DennisAlund: Fixed in 04f6e51, using your suggested phrasing for both.

---

sdk/python/src/shrtnr/_base.py
DennisAlund: Fixed in f68b080, using your suggested phrasing.

---

src/client.ts
DennisAlund: Fixed in 04f6e51: `setLanguage` now checks `res.ok` the same way, with a vitest case for each outcome next to the theme tests.

---

sdk/python/src/shrtnr/_base.py
DennisAlund: Filed as #61 so it doesn't get lost.

---

